PR #1, opened 2020-03-01
Fix some tests in the server

floogulinc:
These tests could have passed when they shouldn't.


floogulinc:
It definitely makes sense to test cases where no user meets the filter criteria. Our server should respond appropriately, I think it returns an empty json lost right now so we should check that is the case and it does not produce an exception for example.

PR #2, opened 2020-03-01
Add Docker

floogulinc:


client/Caddyfile
```diff
@@ -0,0 +1,31 @@
...
+
+    # Rewrite paths that don't start with /api to / for the Angular app
+    rewrite {
+      if {path} not_match ^/api
+      to {path} {path}/ /
+    }
+
+    expires {
+        match .js$ 1y
+        match .css$ 1y
```
NicMcPhee: Does this mean that JS and CSS will be cached for a _year_? I can see how that might make sense for a more stable product, but that doesn't seem good for us. Or am I just misunderstanding something?

server/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+EXPOSE 4567
+# Copy the distribution tar that was built in the build stage
+COPY --from=build /tmp/server/build/distributions/server.tar /app/
+WORKDIR /app
+# Untar the server files
+RUN tar -xvf server.tar
+
+## Add the wait script to the image
+ADD https://github.com/ufoscout/docker-compose-wait/releases/download/2.7.3/wait /wait
+RUN chmod +x /wait
```
NicMcPhee: What is the wait script and why are we including it?

server/src/main/java/umm3601/Server.java
```diff
@@ -16,7 +16,6 @@
 
   static String appName = "UMM CSci 3601 Lab 4";
```
NicMcPhee: We presumably want to change this, or at least add a comment encouraging them to do this. It would be nice to have a list of all the places where they should add/switch to the name of _their_ project.

setupdo.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
```
NicMcPhee: `setupdo` all run together isn't easy to mentally parse. I'd vote instead for something like `setup_droplet.sh`?

server/src/main/java/umm3601/Server.java
NicMcPhee: I just made an issue (#3) for adding the list of places where the project name appears.

client/Caddyfile
floogulinc: When the client is built by webpack the js and css files have a hash of their contents added to their filenames so they are practically immutable. This allows us to cache them aggressively because new versions will be different filenames.

server/Dockerfile
floogulinc: Good idea. The wait command is so the server can wait for mongo to actually be ready before starting. If it starts before mongo it spews some errors (though it still works after mongo comes up).

floogulinc: It is explained in `docker-compose.yml` with the `WAIT_HOSTS` key.

```yml
server:
    build: ./server # Location of the Dockerfile for the client
    restart: always
    environment:
      MONGO_ADDR: mongo # hostname of the mongo container
      MONGO_DB: prod
      WAIT_HOSTS: mongo:27017 # wait for mongo to start up before starting the server
    depends_on:
      - mongo
```

client/Caddyfile
NicMcPhee: I didn't know about the renaming trick. I guess with that it doesn't make much difference. It might be good to have a comment in `Caddyfile` that says something to that effect (maybe with a link to some documentation on that) so people aren't confused by that in the future.

server/Dockerfile
NicMcPhee: It would still be nice to have a pointer here to some info on what the `wait` script does and why we need it. Having a comment in a totally different file isn't immediately helpful.

DEPLOYMENT.md
```diff
@@ -0,0 +1,100 @@
...
+business of hosting Virtual Private Servers (VPS), which they have decided to call
+"droplets." The term isn't terribly important, and for the most part just refers
+specifically to the VPSs hosted by DigitalOcean and all the features which that
+entails.
+
+We will be using a tool called Docker to help with the deployment process. 
+Docker is a tool for creating and running "containers."
+Containers allow a developer to package up an application with all of the parts it needs, 
+such as libraries and other dependencies, and deploy it as one package.
+We will use Docker to separate our app into three containers.
```
NicMcPhee:
```suggestion
We will use Docker to separate our app into three containers: 
```

PR #111, opened 2020-06-20
Run Angular Update GH Actions job only on this org

floogulinc:
This prevents it from running on student repos created from it.

I don't know if this is desired functionality but I thought I'd suggest it.

### DIFF
--- a/.github/workflows/ngx-uptodate.yml
+++ b/.github/workflows/ngx-uptodate.yml
@@ -6,6 +6,7 @@ on: # when the action should run. Can also be a CRON or in response to external 
 jobs:
   ngxUptodate:
     runs-on: ubuntu-latest
+    if: github.repository_owner == 'UMM-CSci-3601'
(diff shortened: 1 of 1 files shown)


NicMcPhee (commented):
So I'm not entirely sure what this does, TBH. I get that your change makes it so that "this" only happens in our GitHub organization (cool – didn't realize you could do that), but I don't really understand what "this" is, and therefore why we'd only want it to happen in our org.

Does this automatically update all our angular dependencies on a regular basis? How does this interact with Dependabot? Aren't they doing essentially the same thing?

floogulinc:
It was supposed to run `ng update` which is prefered to just having dependabot update all the libs because `ng update` runs migrations and dependabot or regular `npm` wouldn't. This action never actually worked very well so I removed it in 13063b527034815747a5fe433ee799bfcd6ec920 but forgot to close this PR.

PR #119, opened 2020-06-28
Switch from Protractor to Cypress for end-to-end testing

floogulinc:
This is currently a work-in-progress exploration of potentially switching the system we use for E2E tests. This PR is here for discussion and to keep track.


floogulinc:
I have gotten Cypress working in GH Actions but we can likely make it better and faster by using the official action: https://github.com/cypress-io/github-action

client/package.json
```diff
@@ -33,10 +32,13 @@
     "@angular/cli": "~9.0.3",
     "@angular/compiler-cli": "~9.0.2",
     "@angular/language-service": "~9.0.2",
-    "@types/node": "^13.7.4",
+    "@nrwl/cypress": "^9.4.5",
```
NicMcPhee, line 35: What are the `@nrwl` things and why do we need them? I would have thought we could have just done something like `npx cypress run` to run the Cypress tests?

floogulinc: the `@nrwl/cypress` package is an Angular CLI builder/schematic and is what allows us to have Cypress run with an `ng` command that automatically builds the project and runs it while passing in its local URL as the base URL to Cypress. It also handles re-running the tests not only when the tests themselves change (like Cypress normally does) but also when the application code changes using `ng e2e --watch` (or `ng cypress --watch` if that's what you set your command name to.

client/e2e/support/user-list-po.ts
```diff
@@ -20,13 +20,21 @@ export class UserListPage {
     return cy.get('.user-nav-list .user-list-item');
   }
 
-  clickViewProfile(card: Cypress.Chainable<HTMLElement>) {
+  clickViewProfile(card: Cypress.Chainable<JQuery<HTMLElement>>) {
     // return card.element(by.buttonText('VIEW PROFILE')).click();
-    return card.contains('VIEW PROFILE').click();
+    return card.find('button');
```
floogulinc, line 25: This looks like it finds any button on the card? Isn't that too fragile, why not have it find the specific button as before? Also the function is still called `clickViewProfile` so why are we having it not actually click it anymore?

kklamberty: That function in the page objects should probably be renamed. I thought it was more consistent behavior to have the page objects give us elements of the page rather than clicking on a specific thing. Clicking and selecting were both actions that were giving me troubles a bit in Cypress given our use of material and Cypress's reluctance to allow things to be clicked on that were not the right kind of thing. I might have been grasping at straws here. The test this is a part of was giving me a very hard time. I would not be at all surprised if this could be made cleaner. I was having a hard time getting at the text on the button, but it was not the largest of my troubles, so I think I just did this. And, this button might be clickable by Cypress without struggle. Please feel free to refactor and make changes, especially here.

floogulinc:
I had setup `npm run cy:open` to have the `baseUrl` set.

client/e2e/support/user-list-po.ts
floogulinc: Turns out this one wasn't actually being used. I have updated it and updated the spec to use it. See [`bd82636` (#119)](https://github.com/UMM-CSci-3601/3601-iteration-template/pull/119/commits/bd8263693ebd13d0bce34a26bd895f3397c2d8c1)

kklamberty:
> I had setup `npm run cy:open` to have the `baseUrl` set.

Got it. That seems reasonable, though I do think that it's a little easier to see what is happening by putting it in `cypress.json`, so I have a slight preference for that approach. I didn't realize we could use the command you listed.

floogulinc:
> > I had setup `npm run cy:open` to have the `baseUrl` set.
> 
> Got it. That seems reasonable, though I do think that it's a little easier to see what is happening by putting it in `cypress.json`, so I have a slight preference for that approach. I didn't realize we could use the command you listed.

Yeah, it's no problem. I had thought that putting that in the `cypress.json` would conflict with what the new `ng e2e` setup wanted to do but after looking through its code I believe it will override the `baseUrl` in `cypresss.json` if needed anyway.

client/e2e/support/user-list-po.ts
floogulinc: If you provide `.contains()` with both a selector and text to look for it will find the first element *of that selector* that contains the given text. Otherwise it was getting the `span` inside the button which isn't really clickable.

```ts
card.contains('button', 'VIEW PROFILE').click();
```

client/e2e/support/user-list-po.ts
```diff
@@ -26,15 +26,12 @@ export class UserListPage {
   }
 
   changeView(viewType: 'card' | 'list') {
-    // in the test of changing the view, .check will not work if you look for mat-radio-button
-    // or, basically anything other than :radio and it will never be ready to click,
-    // so you have to use force: true as an option to not wait for it to be ready
-    return cy.get(`:radio[value="${viewType}"]`).check({ force: true }).click({ force: true });
+    return cy.get(`#view-type-radio .mat-radio-button[value="${viewType}"]`).click();
```
floogulinc: The idea here is just that because of how Angular Material wraps the `input` element its hidden in such a way that a user couldn't actually directly interact with it so Cypress doesn't want to let us either. It seems like a better idea to just click the `.mat-radio-button` itself.

kklamberty:
> > I think we want to encourage students to run `ng serve` in one terminal and `npx cypress open` in another so that they are able to explore the selectors and the console more easily. They can probably explore more easily with that approach than they can if we set it up so they are having things pop up, run the tests, and disappear. That was something frustrating about Protractor that Cypress supports much more easily (unless Protractor could have been doing this all along... in which case we should probably have been running _that_ differently). This sounded like a pretty standard thing to put in cypress.json based on the documentation for [Cypress Configuration](https://docs.cypress.io/guides/references/configuration.html#Options).
> 
> If you use `ng e2e --watch` it opens cypress just like `cypress open` does but while also running the dev server.

That seems reasonable. I know that I'm sometimes poking around using `ng serve` and then want to run tests, so I'm glad for the `ng serve` to be separate, but having options is nice. I'm glad to know about the `--watch` option.

kklamberty:
> This doesn't appear to be doing anything.

I'm not sure I understand. It does what the commit message says it does. It's minor stuff, but it adds in a describe about sidenav and does a couple very small things that were not done before to more completely check the navigation.

kklamberty:
Is the main thing that remains here to do the tests on the New User page? That's the only thing I'm seeing.

kklamberty:
I think all of the E2E tests that were in Protractor (plus some new ones) are included in Cypress in this branch now. @floogulinc and @NicMcPhee do we want to leave the Protractor things in for now and merge this branch so that if we find we have to return to Protractor we don't have to put things back? I'll feel a lot more certain about Cypress after we add something that requires logging in for the template (and labs) and have tried that testing. Thoughts? I think this is ready for review now

client/e2e/support/add-user.po.ts
```diff
@@ -0,0 +1,47 @@
...
+    }
+    if (newUser.email) {
+      this.typeInput('emailField', newUser.email);
+    }
+    this.selectMatSelectValue('roleField', newUser.role);
+    return this.addUserButton().click();
+  }
+
+  checkSnackbar(newUserName: string) {
+    cy.get('.mat-simple-snackbar').invoke('text').should('eq', `Added User ${newUserName}`);
```
floogulinc, line 45: It doesn't seem right to have a `should` in the PO.

client/e2e/integration/user-list.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+    page.getUserCards().first().then((card) => {
+      const firstUserName = card.find('.user-card-name').text();
+      const firstUserCompany = card.find('.user-card-company').text();
+
+      // When the view profile button on the first user card is clicked, the URL should have a valid mongo ID
+      page.clickViewProfile(page.getUserCards().first());
+
+      // The URL should contain '/users/' (note the ending slash) and '/users/' should be followed by a mongo ID
+      cy.url()
+        .should('contain', '/users/')
```
NicMcPhee, line 113: Isn't this included in the `match` on the next line? Can't we just get rid of this?

client/e2e/integration/user-list.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+        .should('match', /.*\/users\/[0-9a-fA-F]{24}$/);
+
+      // On this profile page we were sent to, the name and company should be correct
+      cy.get('.user-card-name').first().should('have.text', firstUserName);
```
NicMcPhee, line 117: This requires that we use the same class on both pages, and imply that we'll be using the card IF on the profile page as well. Do we want to make that kind of assumption?

client/e2e/integration/user-list.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+    });
+   });
+
+  it('Should click add user and go to the right URL', () => {
+    // Click on the button for adding a new user
+    page.addUserFAB().click();
+
+    // The URL contains and ends with '/users/new'
+    cy.url()
+      .should('contain', '/users/new')
```
NicMcPhee, line 128: Again, isn't this handled by the next line?

client/e2e/integration/user-list.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+
+      // On this profile page we were sent to, the name and company should be correct
+      cy.get('.user-card-name').first().should('have.text', firstUserName);
+      cy.get('.user-card-company').first().should('have.text', firstUserCompany);
+    });
+   });
+
+  it('Should click add user and go to the right URL', () => {
+    // Click on the button for adding a new user
+    page.addUserFAB().click();
```
NicMcPhee, line 124: `FAB` is totally non-obvious to me in this name.

client/e2e/support/add-user.po.ts
```diff
@@ -0,0 +1,47 @@
+export interface TestUser {
```
NicMcPhee, line 1: Shouldn't an interface like this be in the production code and pulled from there? Otherwise we're basically defining what a `User` object is in (at least) two places.

client/e2e/integration/user-list.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+    page.getUserTitle().should('have.text', 'Users');
+  });
+
+  it('Should type something in the name filter and check that it returned correct elements', () => {
+    // Filter for user 'Lynn Ferguson'
+    cy.get('#user-name-input').type('Lynn Ferguson');
+
+    // All of the user cards should have the name we are filtering by
+    page.getUserCards().each(e => {
+      cy.wrap(e).find('.user-card-name').should('have.text', 'Lynn Ferguson');
```
NicMcPhee, line 21: Directly looking for `.user-card-name` makes our tests dependent (in many places) on that class. Does that suggest that we really want to have a PO or action here that encapsulates that name so it's not repeated everywhere?

client/e2e/support/add-user.po.ts
```diff
@@ -0,0 +1,47 @@
...
+  navigateTo() {
+    return cy.visit('/users/new');
+  }
+
+  getTitle() {
+    return cy.get('.add-user-title');
+  }
+
+  addUserButton() {
+    return cy.contains('button', 'ADD USER');
```
NicMcPhee, line 19: Does this require that `ADD USER` be in all caps? Are we OK with that?

client/e2e/support/add-user.po.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  typeInput(inputId: string, text: string) {
+    const input = cy.get(`#${inputId}`).click();
```
NicMcPhee, line 23: A think a comment here about string interpolation would probably be a Good Idea.

client/e2e/support/add-user.po.ts
```diff
@@ -0,0 +1,47 @@
...
+    input.type(text);
+  }
+
+  selectMatSelectValue(selectID: string, value: string) {
+    const sel = cy.get(`#${selectID}`).click().get(`mat-option[value="${value}"]`).click();
```
NicMcPhee, line 28: This does a _lot_ of stuff and is not at all obvious to read (at least not to me). We almost certainly want a comment or two, and we might want to break it up into multiple (lexical) lines (if not semantic lines).

client/e2e/support/user-list.po.ts
```diff
@@ -0,0 +1,40 @@
+export class UserListPage {
+  navigateTo() {
+    return cy.visit('/users');
+  }
+
+  getUrl() {
```
NicMcPhee, line 6: We probably don't want to duplicate this function definition.

client/e2e/support/user-list.po.ts
```diff
@@ -0,0 +1,40 @@
...
+  getUserTitle() {
+    return cy.get('.user-list-title');
+  }
+
+  getUserCards() {
+    return cy.get('.user-cards-container app-user-card');
+  }
+
+  getUserListItems() {
+    // return element(by.className('user-nav-list')).all(by.className('user-list-item'));
```
NicMcPhee, line 19: These commented out lines (another one just below) should presumably be deleted. Or, if they serve a purpose, that should be explained in a comment.

client/e2e/support/user-list.po.ts
```diff
@@ -0,0 +1,40 @@
...
+  clickViewProfile(card: Cypress.Chainable<JQuery<HTMLElement>>) {
+    // return card.element(by.buttonText('VIEW PROFILE')).click();
+    return card.contains('button', 'VIEW PROFILE').click();
+  }
+
+  changeView(viewType: 'card' | 'list') {
+    return cy.get(`#view-type-radio .mat-radio-button[value="${viewType}"]`).click();
+  }
+
+  selectRole(value: string) {
```
NicMcPhee, line 32: Should this be an OR type with the set of legal values for the `string` here?

client/e2e/support/user-list.po.ts
```diff
@@ -0,0 +1,40 @@
...
+    // This is messy. I'm not sure that Cypress is working well for us here. Might want to find a better way.
+    return cy.get('#user-role-select').click().get(`mat-option[value="${value}"]`).click();
```
NicMcPhee, line 34: It is a bit messy, but it does accurately reflect the steps required to set a drop-down. If there's an easier/cleaner way I'm good with that, but this seems plausible as is.

That said, we could improve the formatting and include some explanation, e.g.,

```javascript
    return 
        // Find and click the drop down
        cy.get('#user-role-select').click()
          // Select and click the desired value from the resulting menu
          .get(`mat-option[value="${value}"]`).click();
```

client/package.json
```diff
@@ -8,8 +8,7 @@
     "test": "ng test",
     "lint": "ng lint",
     "e2e": "ng e2e",
-    "pree2e": "node node_modules/protractor/bin/webdriver-manager -- update --gecko false",
-    "e2eci": "ng e2e --protractor-config=e2e/protractor-ci.conf.js"
+    "cy:open": "cypress open"
```
NicMcPhee, line 11: Do we actually need this?

client/e2e/integration/app.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  it('Should have the correct title', () => {
+    page.getAppTitle().should('contain', 'CSCI 3601 Iteration Template');
+  });
+
+  describe('Sidenav', () => {
+    it('Should be invisible by default', () => {
+      // Before clicking on the button, the sidenav should be hidden
+      page.getSidenav()
+        .should('be.hidden')
```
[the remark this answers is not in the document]
floogulinc: https://stackoverflow.com/questions/17425543/difference-between-hidden-and-notvisible-in-jquery/17426800

client/e2e/integration/app.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+      page.getSidenav();
+      page.getNavLink('Users').click();
+      cy.url().should('match', /.*\/users$/);
+    });
+
+
+    it('Should have a working navigation to "Home"', () => {
+      page.getSidenavButton().click();
+      page.getNavLink('Home').click();
+      cy.url().should('match', /.*\/$/);
```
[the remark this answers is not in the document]
floogulinc: If Cypress provides a reference to its current `baseUrl` we could probably use that.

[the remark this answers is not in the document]
floogulinc: Also this isn't really testing the navigation to "Home" since we are already on that page.

.github/workflows/e2e.yaml
```diff
@@ -20,29 +20,16 @@ jobs:
...
         java-version: 1.8
-
-    - name: Start server and run e2e tests
-      run: cd server && ./gradlew run & cd client && npm run e2eci
+        
+    - name: Run e2e tests
+      uses: cypress-io/github-action@v2
+      with:
+        start: ${{ github.workspace }}/server/gradlew run -p ${{ github.workspace }}/server/
+        command: npm run e2e
```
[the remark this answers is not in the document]
floogulinc: It could very well be `npx ng e2e`. The reason for using a npm script like this is that it allows you to specify additional parameters in the `package.json` which is good if the command is used in multiple places (eg if you had multiple CI scripts) or if you want to use the npm `before` and `after` hooks.

client/angular.json
```diff
@@ -112,11 +112,11 @@
           }
         },
         "e2e": {
-          "builder": "@angular-devkit/build-angular:protractor",
+          "builder": "@nrwl/cypress:cypress",
```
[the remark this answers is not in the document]
floogulinc: - [`@nrwl/cypress`](https://github.com/nrwl/nx/tree/master/packages/cypress) is the [Angular CLI builder](https://angular.io/guide/cli-builder) that simply runs the dev server (the same as `ng serve`) while running Cypress.
- It looks like you can change it to `"cypress"` but then you need to run it like this: `ng run client:cypress`. So yes the default commands like `e2e`, `test`, `serve`, etc. are baked in similar to how `npm install` and `npm start` are still scripts defined in `package.json` but don't need to be ran with `npm run` like other scripts do.

client/cypress.json
```diff
@@ -0,0 +1,12 @@
+{
+  "baseUrl": "http://localhost:4200/",
+  "fileServerFolder": ".",
+  "fixturesFolder": "e2e/fixtures",
```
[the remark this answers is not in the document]
floogulinc: Yes we can change that quite eaisly.

client/e2e/integration/add-user.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+  it('Should enable and disable the add user button', () => {
+    // ADD USER button should be disabled until...
+    page.addUserButton().should('be.disabled');
+    page.typeInput('nameField', 'test');
+    page.addUserButton().should('be.disabled');
+    page.typeInput('ageField', '20');
+    page.addUserButton().should('be.disabled');
+    page.typeInput('emailField', 'invalid');
+    page.addUserButton().should('be.disabled');
+    page.typeInput('emailField', '@example.com');
```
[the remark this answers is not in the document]
floogulinc: `@example.com` alone wouldn't be but I don't think the `type` function actually clears what is already in the field so in this case `invalid` was already typed so we end up with `invalid@example.com` which is a valid address.

client/e2e/integration/user-list.spec.ts
floogulinc: I suppose we could have a PO function that returns `cy.wrap(e).find('.user-card-name')`

client/e2e/integration/user-list.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+      .should('not.contain.text', 'DATAGENE')
+      .should('not.contain.text', 'OHMNET');
+  });
+
+  it('Should type something in the age filter and check that it returned correct elements', () => {
+    // Filter for users of age '27'
+    cy.get('#user-age-input').type('27');
+
+    // Go through each of the cards that are being shown and get the names
+    page.getUserCards().find('.user-card-name')
```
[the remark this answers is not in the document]
floogulinc: I think the issue with that was that the user list doesn't ever display ages.

client/e2e/integration/user-list.spec.ts
floogulinc: It means "floating action button" but I suppose we could go with just "button"

client/e2e/support/add-user.po.ts
floogulinc: The text in the button in the app is literally "ADD USER".

client/e2e/support/add-user.po.ts
floogulinc: To add to that, why are we setting it as a `const` value and not returning the result?

client/package.json
floogulinc: Probably not if you plan on having students use `npx`.

.github/workflows/e2e.yaml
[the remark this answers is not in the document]
NicMcPhee: Are we using that feature here? Do we think we're likely to? If it's clearly useful, then we can certainly use `npm run` instead, but if we're not likely to to then why not use `npx ng` instead. Can we include comments in the YAML file (not sure if the syntax allows that) to "remind" people that they can switch from `npx ng` to `npm run` if we want/need to put things in `package.json`?

client/angular.json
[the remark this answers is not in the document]
NicMcPhee: Right – thanks. I felt like we'd had this conversation about `@nrwl`, but couldn't remember the answer. That might suggest that a comment of some sort would be good, but that brings us back to the problem of JSON files technically not allowing comments. Sighz.

There's a part of me that would minorly prefer `ng cypress` just to help the students not go searching `ng e2e` and ending up in a bunch of confusing (to them) Protractor documentation. That said, `ng run client:cypress` is kinda ugly by comparison, so maybe we let it go. @kklamberty – what are your thoughts on that?

client/e2e/integration/add-user.spec.ts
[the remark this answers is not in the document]
NicMcPhee: Ah – I hadn't thought about the fact that `invalid` would still be in the field.

I think I'd prefer either (a) clearing the field and typing a "valid" email, or (b) adding a comment above the second `typeInput` reminding the reader that `invalid` would still be in the field, and that we'll end up with `invalid@example.com`. Having a "valid" email with the word "invalid" in it seems at least odd, so I think I'd favor clearing and entering a new email.

client/e2e/integration/app.spec.ts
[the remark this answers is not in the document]
NicMcPhee: That was a long discussion that I didn't find terribly enlightening. If I understood what it was saying, `be.hidden` and `not.be.visible` are always (or often enough) the same, so there's really no point in having both of them in the test, is there? Or am I missing something?

client/e2e/integration/user-list.spec.ts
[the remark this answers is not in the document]
NicMcPhee: Ah, and since this is all E2E, if it's not visible in the UI we can't talk about it. So checking names may be the best thing we can do without changing the UI. It seems like we'd want a comment to that effect?

client/e2e/integration/user-list.spec.ts
NicMcPhee: I'd vote for that, unless there's meaning in "floating action button" that we want/need to capture in some way.

client/e2e/support/add-user.po.ts
NicMcPhee: But the test will fail if we change it to "Add User", which seems awfully fragile to me. Maybe change the test to be case insensitive?

client/e2e/integration/user-list.spec.ts
floogulinc: Its mainly a term specific to Material Design, which in this case is mostly a styling specific thing and functionality wise its the same as any other button.

client/e2e/support/add-user.po.ts
floogulinc: [It looks like you can tell `contains` to be case insensitive](https://docs.cypress.io/api/commands/contains.html#Arguments), so we could certainly do that.

client/e2e/integration/add-user.spec.ts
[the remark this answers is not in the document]
floogulinc: I have implemented clearing the input for `page.typeInput()` and it now types a full email address.

client/e2e/integration/add-user.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+    // all the required fields have valid input, then it should be enabled
+    page.addUserButton().should('be.enabled');
+  });
+
+  it('Should add a new user, go to the right page, and have the right info', () => {
+    const user: TestUser = {
+      name: E2EUtil.randomText(10),
```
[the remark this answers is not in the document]
floogulinc: This has been removed now that we can seed the database fresh for each test we don't need to create unique users every time.

client/e2e/support/user-list.po.ts
floogulinc: I have switched this to using the `UserRole` type from the app.

client/e2e/support/e2e.util.ts
```diff
@@ -0,0 +1,17 @@
+export class E2EUtil {
+  // from https://stackoverflow.com/questions/1349404/generate-random-string-characters-in-javascript
+  public static randomText(length: number): string {
+    let text = '';
+    const possible = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
```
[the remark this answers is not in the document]
floogulinc: This has been removed!

client/e2e/support/add-user.po.ts
floogulinc: This has been done now.

client/e2e/support/add-user.po.ts
floogulinc: This has been fixed by using an attribute to get the button directly.

client/cypress/support/index.ts
```diff
@@ -0,0 +1,20 @@
...
+//
+// You can read more here:
+// https://on.cypress.io/configuration
+// ***********************************************************
+
+// Import commands.js using ES2015 syntax:
+import './commands';
+
+// Alternatively you can use CommonJS syntax:
+// require('./commands')
```
helloworld12321, line 20: Given that `supportFile` is set to `false` in `cypress.json`, should we keep this file around?

I checked, and it looks like cypress only runs this file if you have

```json
"supportFile": "cypress/support/index.ts"
```

in `cypress.json`. 🤷🏻‍♂️

client/cypress/plugins/index.ts
```diff
@@ -0,0 +1,30 @@
...
+
+const mongoHost = process.env.MONGO_ADDR || 'localhost';
+const mongoDb = process.env.MONGO_DB || 'dev';
+
+const mongoUri = `mongodb://${mongoHost}/${mongoDb}`;
+const dbSeedDir = '../database/seed';
+
+const pluginConfig: Cypress.PluginConfig = (on, config) => {
+  // `on` is used to hook into various events Cypress emits
+  // `config` is the resolved Cypress config
```
helloworld12321, line 23: Thanks for adding these comments—that's very helpful.

client/cypress/integration/add-user.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+    page.typeInput(cy.get('#nameField'), 'test');
+    page.addUserButton().should('be.disabled');
+    page.typeInput(cy.get('#ageField'), '20');
+    page.addUserButton().should('be.disabled');
+    page.typeInput(cy.get('#emailField'), 'invalid');
+    page.addUserButton().should('be.disabled');
+    page.typeInput(cy.get('#emailField'), 'user@example.com');
+    // all the required fields have valid input, then it should be enabled
+    page.addUserButton().should('be.enabled');
+  });
```
helloworld12321, line 32: I like this test—it's nice and concise.

client/cypress/integration/app.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  it('Should have the correct title', () => {
+    page.getAppTitle().should('contain', 'CSCI 3601 Iteration Template');
+  });
+
+  it('The sidenav should open, navigate to "Users" and back to "Home"', () => {
+    // Before clicking on the button, the sidenav should be hidden
+    page.getSidenav()
+      .should('be.hidden')
+      .and('not.be.visible');
```
helloworld12321, line 16: Is there a semantic difference between `be.hidden` and `not.be.visible`?

client/cypress/integration/app.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    page.getNavLink('Users').click();
+    cy.url().should('match', /.*\/users$/);
+    page.getSidenav()
+      .should('be.hidden')
+      .and('not.be.visible');
+
+    page.getSidenavButton().click();
+    page.getNavLink('Home').click();
+    cy.url().should('match', /.*\/$/);
```
helloworld12321, line 30: What about a regex like this…?

```suggestion
    cy.url().should('match', /^https?:\/\/[^\/]+\/$/);
```

That way, it should only match the home page, not some other URL that ends with a slash. (Altough, I don't know, maybe the home page is the only URL that ends with a slash ¯\\\_(ツ)\_/¯)

client/cypress/integration/user-list.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+  });
+
+  it('Should type something in the name filter and check that it returned correct elements', () => {
+    // Filter for user 'Lynn Ferguson'
+    cy.get('#user-name-input').type('Lynn Ferguson');
+
+    // All of the user cards should have the name we are filtering by
+    page.getUserCards().each(e => {
+      cy.wrap(e).find('.user-card-name').should('have.text', 'Lynn Ferguson');
+    });
```
helloworld12321, line 28: What do you think about some dollar signs here (just for consistency)?

```suggestion
    page.getUserCards().each($el => {
      cy.wrap($el).find('.user-card-name').should('have.text', 'Lynn Ferguson');
    });
```

client/cypress/integration/user-list.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+    page.getUserCards().should('have.lengthOf.above', 0);
+
+    // Go through each of the cards that are being shown and get the companies
+    page.getUserCards().find('.user-card-company')
+      // We should see these companies
+      .should('contain.text', 'MOMENTIA')
+      .should('contain.text', 'KINETICUT')
+      // We shouldn't see these companies
+      .should('not.contain.text', 'DATAGENE')
+      .should('not.contain.text', 'OHMNET');
```
helloworld12321, line 61: Nice!

client/cypress/integration/user-list.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+      const firstUserName = card.find('.user-card-name').text();
+      const firstUserCompany = card.find('.user-card-company').text();
+
+      // When the view profile button on the first user card is clicked, the URL should have a valid mongo ID
+      page.clickViewProfile(page.getUserCards().first());
+
+      // The URL should contain '/users/' (note the ending slash) and '/users/' should be followed by a mongo ID
+      cy.url()
+        .should('contain', '/users/')
+        .should('match', /.*\/users\/[0-9a-fA-F]{24}$/);
```
helloworld12321, line 126: Since we're just doing a partial string match, we probably don't need the .* at the beginning.

```suggestion
        .should('match', /\/users\/[0-9a-fA-F]{24}$/);
```

client/cypress/integration/user-list.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+    // Click on the button for adding a new user
+    page.addUserButton().click();
+
+    // The URL contains and ends with '/users/new'
+    cy.url()
+      .should('contain', '/users/new')
+      .should(url => expect(url.endsWith('/users/new')).to.be.true);
+
+    // On the page we were sent to, We should see the right title
+    cy.get('.add-user-title').should('have.text', 'New User');
```
helloworld12321, line 144: I might suggest omitting this part of the test—it seems like a bit of a duplicate of what we have in `add-user.spec.ts`.

```suggestion
```

client/cypress/integration/add-user.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+    beforeEach(() => {
+      cy.task('seed:database');
+    });
+
+    it('Should go to the right page, and have the right info', () => {
+      page.addUser(user);
+
+      cy.url()
+        .should('match', /.*\/users\/[0-9a-fA-F]{24}$/)
+        .should('not.match', /.*\/users\/new$/);
```
helloworld12321, line 54:
```suggestion
        .should('match', /\/users\/[0-9a-fA-F]{24}$/)
        .should('not.match', /\/users\/new$/);
```

client/cypress/integration/app.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+    page.getSidenav()
+      .should('be.hidden')
+      .and('not.be.visible');
+
+    page.getSidenavButton().click()
+      .should('not.be.hidden')
+      .and('be.visible');
+
+    page.getNavLink('Users').click();
+    cy.url().should('match', /.*\/users$/);
```
helloworld12321, line 23:
```suggestion
    cy.url().should('match', /\/users$/);
```

client/cypress/integration/app.spec.ts
helloworld12321: (If the latter, I might recommend this change instead. Up to you! 🙂)

```suggestion
    cy.url().should('match', /\/$/);
```

helloworld12321 (changes requested):
This is really cool!

I've made a few minor suggestions, and then one or two questions about the implementation.

Thanks for doing all this, Paul + KK + Nic! :D

client/cypress/integration/add-user.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+describe('Add user', () => {
+  const page = new AddUserPage();
+
+  beforeEach(() => {
+    page.navigateTo();
+  });
+
+  after(() => {
+    // Lets put the db back to its original state before we leave.
+    //cy.task('seed:database');
```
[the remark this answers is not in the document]
floogulinc: I wasn't sure if we wanted that or not. One one hand, Cypress recommends against resettings things in "after" steps because it makes it harder to inspect things when something goes wrong but at the same time I think it's a good idea to make sure the tests end with the db seeded back to normal since its the same one students use for development.

client/cypress/integration/add-user.spec.ts
floogulinc: I feel like we could be more specific here and actually try to enforce that the _path_ should be exactly `/users\/[0-9a-fA-F]{24}$` but without doing that this is better than before.

client/cypress/integration/app.spec.ts
floogulinc: We had a whole discussion about that above. It does seem like there is a difference: https://stackoverflow.com/questions/17425543/difference-between-hidden-and-notvisible-in-jquery/17426800

client/cypress/integration/app.spec.ts
floogulinc: We should really capture both localhost:4200 with and without a trailing slash.

client/cypress/integration/user-list.spec.ts
floogulinc: I'm personally in favor of using the dollar sign when it's referring to a JQuery object but @NicMcPhee didn't seem to like doing that so maybe we actually remove it from the ones that currently have it.

client/cypress/integration/user-list.spec.ts
floogulinc: I like having a second check that we are indeed at the right page personally.

client/cypress/plugins/index.ts
floogulinc: Haha. they are just part of the default in this file!

client/cypress/support/index.ts
floogulinc: Oh then we should probably fix `cypress.json`. I think we should have these default files available incase someone wants to use custom commands.

client/cypress/integration/app.spec.ts
floogulinc:
```suggestion
    cy.url().should('match', /^https?:\/\/[^\/]+\/?$/);
```
Would this achieve that @helloworld12321?

client/cypress/integration/add-user.spec.ts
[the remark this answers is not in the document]
helloworld12321: Yeah, I suppose the underlying problem is that it would be preferable to use the `test` database instead of the `dev` database for this. (But we would need a way to communicate that to the server somehow. 🙁)

My first preference would be to remove the `after(() => {})` block entirely, but I'd also be fine with un-commenting the call to `cy.task('seed:database')`. Do you have a preference? 🙂

client/cypress/support/index.ts
helloworld12321: Oh, okay! 🙂 I'll go make a suggestion to that effect.

client/cypress.json
```diff
@@ -0,0 +1,12 @@
+{
+  "baseUrl": "http://localhost:4200/",
+  "fileServerFolder": ".",
+  "fixturesFolder": "cypress/fixtures",
+  "integrationFolder": "cypress/integration",
+  "pluginsFile": "cypress/plugins/index.ts",
+  "supportFile": false,
```
helloworld12321, line 7: See this earlier thread: https://github.com/UMM-CSci-3601/3601-iteration-template/pull/119#discussion_r579720794

```suggestion
  "supportFile": "cypress/support/index.ts",
```

client/cypress/integration/app.spec.ts
helloworld12321: Neat, thank you!

In that case, would it be all right to only include one of these two assertions? (Just to avoid redundancy.)

client/cypress/integration/app.spec.ts
helloworld12321: Nice, looks good! 🙂

client/cypress/integration/user-list.spec.ts
helloworld12321: Sounds good, that works too! :)

NicMcPhee (approved):
This looks great! Thanks for all your work on it, @floogulinc!

I've proposed a few commit suggestions, but they're all in comments and none are crucial IMO. Still, if we can bring them in and then merge that would be cool.

And thanks to @helloworld12321 for the tip about being able to only review new changes – made this _way_ more doable. 😜

NicMcPhee (approved):
This looks nice – thanks!

helloworld12321 (approved):
Looks good! Thanks so much @floogulinc for all of your hard work :)

helloworld12321 (approved):

PR #136, opened 2020-07-17
Update to Angular 10

### LINKED ISSUE #118: Update to Angular 10
See https://blog.angular.io/version-10-of-angular-now-available-78960babd41

### LINKED ISSUE #138: Update Angular Material and flex-layout
From @floogulinc [here](https://github.com/UMM-CSci-3601/3601-iteration-template/pull/136#issuecomment-660432107):

> You also need to update Angular Material and flex-layout. That should be: `ng upgrade @angular/material @angular/flex-layout`

NicMcPhee:
@floogulinc pointed out that Angular has upgraded to v10 and that we should update (#118); this PR updates the project to Angular 10 [using the update guide](https://update.angular.io/#9.0:10.0).

I also (by hand) converted the project to the strictness settings that ng new --strict would provide. In order to make sure that hadn't broken anything, I (by hand) fixed the broken tests.

Closes #118 
Closes #138 

I'm also assuming that this will close or make obsolete a number of Dependabot pull requests that essentially come down to updating various Angular libraries and dependencies.


floogulinc:
You also need to update Angular Material and flex-layout. That should be: `ng upgrade @angular/material @angular/flex-layout`

floogulinc:
Also we should probably move the tsconfig files around to match the new defaults.

NicMcPhee:
These changes are identical to those made earlier in [the revised Angular tutorial](https://github.com/UMM-CSci-3601/revised-angular-tutorial/pull/6), so I don't expect there to be any issues. If someone can confirm that it looks OK I'll merge that in, which I think will clear out quite a few of the large pile of Dependabot PRs. (I'm trying to get through a lot of them now so they don't all just pile up until winter break or next Spring.

NicMcPhee:
> You also need to update Angular Material and flex-layout. That should be: `ng upgrade @angular/material @angular/flex-layout`

Why do I need to do those as part of *this* PR? I recognize that those updates should happen, but I'd like this to not turn into the "kitchen sink" of "update everything that's out of date". (As the Dependabot PRs indicate, there's a _lot_ that's out of date.) We didn't update those in the updates in the [BDD version of the tutorial](https://github.com/UMM-CSci-3601/revised-angular-tutorial/pull/6); that probably needs to happen over there as well?

So my inclination would be to make a new issue for upgrading those if Dependabot hasn't already done so.

> Also we should probably move the tsconfig files around to match the new defaults.

What needs to happen here? The migration guide suggests that at least some of this was automagically done by `ng update`, but presumably there's more that should happen?

client/package.json
```diff
@@ -8,6 +8,7 @@
     "test": "ng test",
     "lint": "ng lint",
     "e2e": "ng e2e",
+    "postinstall": "ngcc",
```
floogulinc, line 11: I think this actually slows down our CI builds because with this I think it has to generate all possible package formats for all libraries in `node_modules` whereas `ng serve` and `ng build` will run `ngcc` as needed automatically.

floogulinc:
> Why do I need to do those as part of _this_ PR? I recognize that those updates should happen, but I'd like this to not turn into the "kitchen sink" of "update everything that's out of date". (As the Dependabot PRs indicate, there's a _lot_ that's out of date.) We didn't update those in the updates in the [BDD version of the tutorial](https://github.com/UMM-CSci-3601/revised-angular-tutorial/pull/6); that probably needs to happen over there as well?
> 
> So my inclination would be to make a new issue for upgrading those if Dependabot hasn't already done so.

The tutorial doesn't use Angular Material or flex-layout so that's why they aren't updated there. 

Generally Angular Material is considered a core part of Angular and tracks the same versions as the main library. Note that on the Angular update guide there is literally a "I use Angular Material" checkbox that adds "Run `ng update @angular/material`." to the list of steps. Additionally you will want to update `@angular/flex-layout` because it depends on Angular CDK as a peer dependency and Angular CDK is updated when Angular Material will. Updating Angular Material without updating flex-layout will fail because it would result in the flex-layout peer dependency not being met.

Additionally, there appears to be an update to `rxjs` that could be applied with `ng update rxjs` 


Also, in the future when running ng update, the `--createCommits=true` parameter is nice since it will create commits for each migration step.

floogulinc:
> What needs to happen here? The migration guide suggests that at least some of this was automagically done by `ng update`, but presumably there's more that should happen?

I think i was mistaken about this. It looks like it was migrated properly.

client/package.json
floogulinc: I suppose that can be mitigated if we have GH Actions cache the compiled modules.

NicMcPhee:
> Note that on the Angular update guide there is literally a "I use Angular Material" checkbox that adds "Run `ng update @angular/material`." to the list of steps

Duh – not sure how I missed that. I'll update those as part of this, then.

> Additionally, there appears to be an update to `rxjs` that could be applied with `ng update rxjs`

Cool.
 
> Also, in the future when running ng update, the `--createCommits=true` parameter is nice since it will create commits for each migration step.

I think you even mentioned that somewhere before, but I'm not sure where. How would you see that being concretely useful? Are you thinking it will just make it easier to see what got changed? Or that we might actually cherry pick commits from one project to another?

I'll do the material and rxjs updates here in a bit, and then re-request the review.

floogulinc:
> I think you even mentioned that somewhere before, but I'm not sure where. How would you see that being concretely useful? Are you thinking it will just make it easier to see what got changed? Or that we might actually cherry pick commits from one project to another?

I think it mainly nice because it makes decent commits explaining each step of the migration rather than "I ran the upgrade script."

Here are some examples from when I updated one of my side projects to Angular 10:

image

Each commit has a detailed description as well, for example:


> ### @angular/cli migration - solution-style-tsconfig
>Add "Solution Style" TypeScript configuration file support. This improves developer experience using editors powered by TypeScript’s language server. Read more about this here: https://v10.angular.io/guide/migration-solution-style-tsconfig

and

> ### @angular/cli migration - remove-es5-browser-support-option
>Remove deprecated 'es5BrowserSupport' browser builder option. The inclusion for ES5 polyfills will be determined from the browsers listed in the browserslist configuration.

client/package.json
NicMcPhee: Do you think we should just remove this altogether? I haven't looked into this in any details; what they say in the upgrade guide is

> If you depend on many Angular libraries you may consider speeding up your build by invoking the ngcc (Angular Compatibility Compiler) in an npm postinstall script via small change to your package.json.

I wasn't clear on what constituted "many Angular libraries", but it seemed like we depended on quite a few so I thought this might improve performance for students. Are you saying that `ng serve` will automagically call `ngcc` so we don't need this in the `package.json`?

NicMcPhee: After updating Material and rxjs I get this warning:

> Warning: The inferred tsconfig file "/Users/mcphee/Downloads/3601-iteration-template/client/tsconfig.json" appears to be "solution-style" since it contains no root files but does contain project references.
This is probably not wanted, since ngcc is unable to infer settings like "paths" mappings from such a file.
Perhaps you should have explicitly specified one of the referenced projects using the --tsconfig option. For example:

>  ngcc ... --tsconfig "./tsconfig.app.json"
>  ngcc ... --tsconfig "./tsconfig.spec.json"

> Find out more about solution-style tsconfig at https://devblogs.microsoft.com/typescript/announcing-typescript-3-9/#solution-style-tsconfig.
> If you did intend to use this file, then you can hide this warning by providing it explicitly:

>  ngcc ... --tsconfig "tsconfig.json"
> removed 1 package and audited 1631 packages in 10.374s

I didn't get this on the other project, so I'm not really sure what triggered it. Everything still seems to work (e.g., tests pass), but this does suggest that best we don't have things set up "best" for `ngcc`.

Should I just remove this line?

floogulinc:
Superseded by #306

Issue #53, opened 2020-04-24
Improve ActivatedRouteStub observable behavior


floogulinc:
This is the current `ActivatedRouteStub` from [here](https://angular.io/guide/testing#activatedroutestub) in the Angular Docs:

```ts
import { convertToParamMap, ParamMap, Params } from '@angular/router';
import { ReplaySubject } from 'rxjs';

/**
 * An ActivateRoute test double with a `paramMap` observable.
 * Use the `setParamMap()` method to add the next `paramMap` value.
 */
export class ActivatedRouteStub {
  // Use a ReplaySubject to share previous values with subscribers
  // and pump new values into the `paramMap` observable
  private subject = new ReplaySubject<ParamMap>();

  constructor(initialParams?: Params) {
    this.setParamMap(initialParams);
  }

  /** The mock paramMap observable */
  readonly paramMap = this.subject.asObservable();

  /** Set the paramMap observables's next value */
  setParamMap(params?: Params) {
    this.subject.next(convertToParamMap(params));
  }
}
```

One of the issues we have run into is that the behavior of the `ReplaySubject` of replaying every value that has been sent to it to all subscribers seems like the wrong thing to do for stubbing the current route params. It seems like it would be better if it only replayed the last value it was sent. To do this we can simply pass in a `bufferSize` parameter of `1` in the constructor of the `ReplaySubject`:

```ts
  private subject = new ReplaySubject<ParamMap>(1);
```

Additionally the behavior of the constructor can be somewhat confusing. It is not obvious that not providing a value in the constructor results in `null` being the emitted to subscribers of `paramMap`. We should add proper JSDoc documentation and maybe change the constructor's implementation.

NicMcPhee:
We should make sure we indicate in the docs that we're modifying the code in the Angular Docs and why; otherwise students who are looking at theirs and ours are going to be pretty confused.

I like the idea of specifying the `bufferSize` – that should definitely help lead to more predictable/understandable behaviors.

I agree that the constructor seems problematic. One option would be to make `initialParams` no longer optional, so people can't construct it with that `null` value implicitly in the pipeline. The problem there is that there could be cases where we want _one_ instance of this stub that we keep reusing by calling `setParamMap` to change the `params`. In that scenario it probably doesn't make sense to force the user to specify a value when the stub is first constructed.

Another idea that I've thought about but never pursued would be to follow up on their suggestion right after this code:

[...]

PR #137, opened 2020-07-18
Turn on Angular 10 strictness

NicMcPhee:
In Angular 10, `ng new` has a new `--strict` option that turns on a host of strictness checks for your new project. I would like to turn those on for this project as well (3c341fa), but doing so leads to a whole lot of errors in our code. This PR will fix all those and get us back to clean and tests passing.


floogulinc:
You also need to update Angular Material and flex-layout

client/src/testing/activated-route-stub.ts
```diff
@@ -19,8 +19,10 @@ export class ActivatedRouteStub {
   /** The mock paramMap observable */
   readonly paramMap = this.subject.asObservable();
 
-  /** Set the paramMap observables's next value */
+  /** Set the paramMap to the next value of the observable */
   setParamMap(params?: Params) {
```
floogulinc, line 23: I wonder why this had an optional param in the first place, it seems like this could be simplified by making it a required param and removing the check

client/src/app/users/user-list.component.ts
```diff
@@ -12,25 +12,23 @@ import { Subscription } from 'rxjs';
...
+  public filteredUsers: User[] = [];
 
-  public userName: string;
-  public userAge: number;
-  public userRole: UserRole;
-  public userCompany: string;
+  public userName: string = '';
+  public userAge: number = -1;
+  public userRole: UserRole = 'viewer';
+  public userCompany: string = '';
```
floogulinc, line 21: I don't think we want to set defaults for the filters like this.

client/src/app/users/user-card.component.ts
```diff
@@ -8,8 +8,8 @@ import { User } from './user';
 })
 export class UserCardComponent implements OnInit {
 
-  @Input() user: User;
-  @Input() simple ? = false;
+  @Input() user?: User;
```
floogulinc, line 11: Why make this optional?

client/src/app/users/user-list.component.ts
NicMcPhee: I didn't like that either, but I wasn't sure how to avoid it. These essentially become the "default" values of the filter fields in the user list component.

A lot of these changes were really rushed. I initially thought it would be "easy" to fix these things and I was trying to include it in the Angular 10 update (#136). After a while, though, I realized that I was typing more than thinking and decided to split this off, so we could look it over and have exactly this sort of conversation.

So what do you think would be a better approach? Make them all optional? That seems to be an option, although I'm nervous about just making a bazillion things optional as it seems sloppy/lazy to me.

client/src/testing/activated-route-stub.ts
NicMcPhee: It's a slightly odd design. The constructor takes that as an optional argument, and then calls `setParamMap()` with that (optional) value. I presume that's why it's marked as optional, because it inherits that from the constructor. The way things are _used_ is that the constructor _never_ gets an argument, but `setParamMap` _always_ gets one.

A plausible option would be to make the `setParamMap` argument required, and have the constructor check to see if it was actually given a value and only call `setParamMap` if it was.

client/src/app/users/user-card.component.ts
NicMcPhee: There can't really be an initial value for it here since it's value is coming from somewhere else (hence the `@Input`). The Internet seems to suggest that `?` is the way to go here, but I confess to not being thrilled about this decision. Do you have another suggestion?

client/src/testing/activated-route-stub.ts
floogulinc: I think this relates to discussion in #53

client/src/app/users/user-list.component.ts
floogulinc: Is there something preventing them from just being uninitialized?

NicMcPhee: Yup. With all the strictness stuff turned on you get errors because it can't be sure that they have values when they're used. This may indicate a weakness in the overall design of this component?

floogulinc:
I'm not sure we want to do this anymore but if we do it will need to be in a new PR anyway so I'm closing this.

PR #255, opened 2021-01-07
Update client to use Caddy 2

### LINKED ISSUE #18: Update to Caddy 2
There is a new major version of Caddy still in beta. It looks like it will be out of beta soon and we should be ready to switch to it.

floogulinc:
Closes #18


NicMcPhee:
@floogulinc – Is this done and actually waiting for me to review it? Should I run through the process on a droplet and see if there are any surprises?

floogulinc:
> @floogulinc – Is this done and actually waiting for me to review it? Should I run through the process on a droplet and see if there are any surprises?

No, I will request your review when it is ready.

floogulinc:
@NicMcPhee I think I have it fixed. I have removed the option to use plain http because the `auto_https off` directive just wasn't working and we probably don't want to encourage using http anyway.

setupdroplet.sh
```diff
@@ -12,44 +12,31 @@ else
...
-  echo "http://${domain}"
+echo "as well as the ZeroSSL Terms of Service at:"
+echo "https://zerossl.com/terms/"
+echo
+echo "Please enter your email address to signify agreement and to be notified"
+echo "in case of issues."
+read -p "Email address: " email
+echo
+if [ -z "$email" ]; then
+  echo "Not setting APP_CADDY_GLOBAL_OPTIONS"
```
NicMcPhee, line 33:
```suggestion
  echo "No email entered; not setting APP_CADDY_GLOBAL_OPTIONS"
```

NicMcPhee:
I just went through this, and it appeared to work fine. Some comments:

* It would be good to update the Deployment instructions a little to reflect:
  * The inclusion of ZeroSSL
  * The need to enter an email (and why/how that will be used)
* There's a warning that we might want to look at?
  * `WARNING: Image for service server was built because it did not already exist. To rebuild this image you must use 'docker-compose build' or 'docker-compose up --build'.`
  * This comes right after successfully building and tagging the server.
  * And the same thing happens again after building and tagging the client.
* There's still that big pile of red when it tries to do `node-gyp rebuild`.

The instructions update should probably happen before we merge this in. I'm happy if we create issues for the `node-gyp rebuild` errors and (if we think it's worth it) the warning about the image so we can merge this in.

NicMcPhee (approved):
This looks great. We've created an issue for the `node-gyp` error and we're not going to worry about the warnings. The changes to the instructions will move to that pull request.

PR #256, opened 2021-01-08
Update Java, Gradle, and almost all the libraries

NicMcPhee:
This updates a number of things on the server side:

* Gradle to v6.8
* Java to JDK 11
* Almost all the Java dependencies

Gradle 6.8 was only _just_ release (literally in the last 24-48 hours), so we may want to re-update for minor fixes and the like closer to class.

This uses the new-ish Gradle toolchain system to encode the dependency on Java 11. We used to use the `sourceCompatibility` and `targetCompatibility` variables to indicate which version of Java we wanted Gradle to use. The new toolchain option will actually download the specified JDK from AdoptOpenJDK if it's not already available on the system. That's a bit slow the first time you run Gradle (since it's installing a full JDK), but is fast enough after that, and it may simplify some of the occasional issues of JDK mismatches.

All the Java dependencies are updated to their latest version except for the Jackson dependencies. Attempting to upgrade either `mongojack` or `jackson-databind` breaks the code (things don't even compile), so some non-trivial work will need to be done to fix that.

All the tests pass locally, and everything passes in GitHub Actions as well.

Closes #244 
Closes #226 
Closes #214 
Closes #188 
Closes #187 
Closes #162


.github/workflows/e2e.yaml
```diff
@@ -37,11 +37,6 @@ jobs:
       run: npm ci
       working-directory: ./client
 
-    - name: Set up JDK 11
```
NicMcPhee, line 40: This section is remove here and in the `server-gradle.yml` because the appropriate JDK is now downloaded by Gradle if necessary.

server/build.gradle
```diff
@@ -13,6 +13,12 @@ plugins {
   id 'jacoco'
 }
 
+java {
```
NicMcPhee, line 16: This is the new toolchain mechanism for specifying which version of Java we want to use.

server/build.gradle
```diff
@@ -26,19 +32,21 @@ mainClassName = 'umm3601.Server'
...
-  implementation 'io.javalin:javalin:3.7.0'
+  implementation 'io.javalin:javalin:3.12.0'
 
   // Mongo DB Driver for Java
-  implementation 'org.mongodb:mongodb-driver:3.12.5'
+  implementation 'org.mongodb:mongodb-driver:3.12.7'
 
   // Jackson, a JSON library for Java
   implementation 'com.fasterxml.jackson.core:jackson-databind:2.10.2'
+  // implementation 'com.fasterxml.jackson.core:jackson-databind:2.12.0'
```
NicMcPhee, line 45: This and the next line are my failed attempts to upgrade the Jackson libraries. If we accept this PR I'll probably delete those comments before merging in, and instead create new GitHub issues about these upgrades.

server/build.gradle
```diff
@@ -49,29 +57,24 @@ dependencies {
...
+  testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.7.0'
 
-  testImplementation 'com.mockrunner:mockrunner-servlet:2.0.1'
+  testImplementation 'com.mockrunner:mockrunner-servlet:2.0.6'
 }
 
 application {
   // Define the main class for the application
-  mainClassName = 'umm3601.Server'
+  mainClass.set("umm3601.Server")
```
NicMcPhee, line 70: This changes a now deprecated approach to setting the main class name to the "new improved" approach.

server/build.gradle
```diff
@@ -49,29 +57,24 @@ dependencies {
...
 }
 
 test {
   // Use junit platform for unit tests
   useJUnitPlatform()
-
 }
 
-// We want our source code to be compatible with Java 11
```
NicMcPhee, line 71: We don't need this anymore because we're using the toolchain system.

.github/workflows/e2e.yaml
```diff
@@ -9,7 +9,7 @@ jobs:
 
     services:
       mongo:
-        image: mongo:3.4.23
+        image: mongo:4.4.3
```
floogulinc, line 12: Do we want to just set this to `mongo` without a version like we do in the docker-compose file?

.github/workflows/server-gradle.yml
```diff
@@ -9,7 +9,7 @@ jobs:
 
     services:
       mongo:
-        image: mongo:3.4.23
+        image: mongo:4.4.3
```
floogulinc, line 12: See comment on `e2e.yml`

floogulinc (changes requested):

.github/workflows/e2e.yaml
floogulinc: Any idea where it puts the downloaded JDK so we can cache it?

floogulinc:
Suggestion: modify the gradle cache section of the e2e and Java test GitHub Actions workflows to this:
```yml
- uses: actions/cache@v2
  with:
    path: |
      ~/.gradle/caches
      ~/.gradle/wrapper
      ~/.gradle/jdks
    key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*') }}
    restore-keys: |
      ${{ runner.os }}-gradle-
```

This will have GH Actions cache not just the jars we have been caching but also the Gradle wrapper itself and (potentially most importantly) the JDK it downloads.

.github/workflows/e2e.yaml
floogulinc: Nevermind, I looked it up in the docs. See https://github.com/UMM-CSci-3601/3601-iteration-template/pull/256#issuecomment-757048506

.github/workflows/e2e.yaml
NicMcPhee: Probably, actually. Good catch.

NicMcPhee:
Nice suggestion – I'll do that here in a sec. Assuming that passes, etc., would you be good with merging this @floogulinc?

NicMcPhee:
I added the caching, and everything looks good. I also removed the Mongo version and updated all the actions to v2 since that seems to be the latest version of all of them. If @floogulinc thinks all this looks good we'll be ready to merge!

floogulinc (dismissed):

floogulinc (commented):
Do we still need these lines:

https://github.com/UMM-CSci-3601/3601-iteration-template/blob/5f495dbc2dbf2bd0c277606af6facf2b6708e0a6/server/build.gradle#L29-L30

PR #257, opened 2021-01-09
Update jackson

NicMcPhee:
The versions of the Jackson libraries are substantially behind, and we should update them.

I attempted to do this here, but one (and only one) of the tests don't pass, and I can't for the life of me figure out why. @floogulinc – if you have any ideas on why this isn't working, that would be awesome.

There's a lot of commented out code and other cruft here at the moment that would need to be cleaned up before this can be merged.

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -38,15 +38,15 @@ dependencies {
   implementation 'io.javalin:javalin:3.12.0'
 
   // Mongo DB Driver for Java
-  implementation 'org.mongodb:mongodb-driver:3.12.7'
+  implementation 'org.mongodb:mongo-java-driver:3.12.7'
 
   // Jackson, a JSON library for Java
-  implementation 'com.fasterxml.jackson.core:jackson-databind:2.10.2'
-  // implementation 'com.fasterxml.jackson.core:jackson-databind:2.12.0'
+  // implementation 'com.fasterxml.jackson.core:jackson-databind:2.10.2'
+  implementation 'com.fasterxml.jackson.core:jackson-databind:2.12.0'
 
   // MongoJack, MongoDB integration for Jackson
-  implementation 'org.mongojack:mongojack:2.10.0'
-  // implementation 'org.mongojack:mongojack:4.0.2'
+  // implementation 'org.mongojack:mongojack:2.10.0'
+  implementation 'org.mongojack:mongojack:4.0.2'
 
   // Jackson support modules for Java 8 datatypes
   implementation 'com.fasterxml.jackson.module:jackson-module-parameter-names:2.10.2'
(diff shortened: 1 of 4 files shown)


server/build.gradle
```diff
@@ -32,15 +32,15 @@ dependencies {
   implementation 'io.javalin:javalin:3.12.0'
 
   // Mongo DB Driver for Java
-  implementation 'org.mongodb:mongodb-driver:3.12.7'
+  implementation 'org.mongodb:mongo-java-driver:3.12.7'
```
floogulinc, line 35: These two are not equivalent. Also these particular packages haven't continued through to Mongo 4.

server/build.gradle
```diff
@@ -32,15 +32,15 @@ dependencies {
...
-  implementation 'com.fasterxml.jackson.core:jackson-databind:2.10.2'
-  // implementation 'com.fasterxml.jackson.core:jackson-databind:2.12.0'
+  // implementation 'com.fasterxml.jackson.core:jackson-databind:2.10.2'
+  implementation 'com.fasterxml.jackson.core:jackson-databind:2.12.0'
 
   // MongoJack, MongoDB integration for Jackson
-  implementation 'org.mongojack:mongojack:2.10.0'
-  // implementation 'org.mongojack:mongojack:4.0.2'
+  // implementation 'org.mongojack:mongojack:2.10.0'
+  implementation 'org.mongojack:mongojack:4.0.2'
```
floogulinc, line 43: This version of mongojack requires the MongoDB driver 4.x

We should use
```gradle
implementation 'org.mongodb:mongodb-driver-sync:4.1.1
```

floogulinc, line 43: Coming from mongojack 2.x to 3.x and newer there is a migration guide to follow: https://github.com/mongojack/mongojack/blob/master/MIGRATING.md

server/build.gradle
NicMcPhee: Good catch about the Mongo 4, but that's a whole 'nother set up updates, and looks like it's a non-trivial set of changes that we'd have to make. I'll create an issue about it, but I'm not sure it's happening right now.

server/build.gradle
NicMcPhee: Ah, so we can't upgrade Jackson that far without also upgrading Mongo. I had totally missed that part.

We can't just simply upgrade to either of the 4.1.1 drivers (`-sync` or `-legacy`) as both then break things in a variety of ways. So it does seem that we want/need to upgrade to Mongo 4 first, and then sort out the Jackson upgrade.

server/build.gradle
floogulinc: We are already using Mongo 4 so this is probably necessary.

server/build.gradle
floogulinc: We are already using Mongo 4 at this point so I would go ahead and start using the new driver.

floogulinc:
Superseded by #305

PR #305, opened 2021-02-18
Spring 2021 Server Updates

floogulinc:


server/src/main/java/umm3601/user/UserController.java
```diff
@@ -32,19 +30,16 @@
...
   /**
    * Construct a controller for users.
    *
    * @param database the database containing user data
    */
   public UserController(MongoDatabase database) {
-    jacksonCodecRegistry.addCodecForClass(User.class);
-    userCollection = database.getCollection("users").withDocumentClass(User.class)
-        .withCodecRegistry(jacksonCodecRegistry);
+    userCollection = JacksonMongoCollection.builder().build(database, "users", User.class);
```
helloworld12321, line 42: That's pretty nifty! :)

helloworld12321 (approved):
Looks great! Thanks! :)

server/src/main/java/umm3601/Server.java
```diff
@@ -41,22 +42,19 @@ public static void main(String[] args) {
...
-    // Simple example route
-    server.get("hello", ctx -> ctx.result("Hello World"));
-
     // Utility routes
-    server.get("api", ctx -> ctx.result(appName));
+    server.get("/api", ctx -> ctx.result(appName));
 
     // Get specific user
-    server.get("api/users/:id", userController::getUser);
+    server.get("/api/users/:id", userController::getUser);
```
NicMcPhee, line 49: The order of these paths is kind random and probably reflects history more than anything else. I might re-arrange them and add that to the PR.

NicMcPhee (changes requested):
This looks great. I think I'm going to re-arrange the paths in `Server.java` so they seem a little more coherent, but the rest is nice. I really like the `JacksonMongoCollection` trick – much nicer.

server/src/main/java/umm3601/Server.java
```diff
@@ -32,6 +32,7 @@ public static void main(String[] args) {
         builder.hosts(Arrays.asList(new ServerAddress(mongoAddr))))
       .build());
 
+
```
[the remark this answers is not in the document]
floogulinc: Feel free to remove it.

server/src/main/java/umm3601/Server.java
```diff
@@ -26,39 +24,47 @@ public static void main(String[] args) {
...
     // Initialize dependencies
     UserController userController = new UserController(database);
-    //UserRequestHandler userRequestHandler = new UserRequestHandler(userController);
-
-    Javalin server = Javalin.create().start(4567);
 
-    // Simple example route
-    server.get("hello", ctx -> ctx.result("Hello World"));
+    Javalin server = Javalin.create();
+    /*
```
NicMcPhee, line 40: There were warnings about the fact that our `MongoClient` isn't necessarily closed if something Bad happens and, for example, the server crashes.

The recommendations were to use the try-with-resources or a `try` with a `finally` block, but those don't work here. The `MongoClient` variable has to be "live" the entire time that the server is up, so we can't shut it down anywhere in `main()`.

This adds server event handlers that close the client when either the server fails to actually start, or the server is stopping. I don't know if those are the right ones or enough, so feedback would certainly be appreciated.

I also don't know if this is several more lines of code and additional complication that doesn't actually make anything better, so thoughts there would be good as well.

server/src/main/java/umm3601/user/UserController.java
```diff
@@ -85,19 +83,19 @@ public void deleteUser(Context ctx) {
    */
   public void getUsers(Context ctx) {
 
-    List<Bson> filters = new ArrayList<Bson>(); // start with a blank document
+    List<Bson> filters = new ArrayList<>(); // start with a blank document
 
-    if (ctx.queryParamMap().containsKey("age")) {
-        int targetAge = ctx.queryParam("age", Integer.class).get();
-        filters.add(eq("age", targetAge));
+    if (ctx.queryParamMap().containsKey(AGE_KEY)) {
```
NicMcPhee, line 88: SonarLint fussed about the string "company" appearing three times in `getUsers()`, which seemed like a reasonable complaint. So I converted the three key strings ("age", "company", and "role") into string constants.

One could argue that similar things can/should be done for some of the other string constants as well, but they're not repeated as much so I didn't do that here.

One could also argue that creating these constants is overkill here, so feel free to push back.

server/src/main/java/umm3601/Server.java
NicMcPhee: Oh, I clearly put this on the wrong line. The relevant code is:

```java
Javalin server = Javalin.create();
    /*
     * We want to shut the `mongoClient` down if the server either
     * fails to start, or when it's shutting down for whatever reason.
     * Since the mongClient needs to be available throughout the
     * life of the server, the only way to do this is to wait for
     * these events and close it then.
     */
    server.events(event -> {
      event.serverStartFailed(mongoClient::close);
      event.serverStopping(mongoClient::close);
    });
    server.start(4567);
```

NicMcPhee (approved):

server/src/main/java/umm3601/Server.java
helloworld12321: Looks reasonable!

It seems like like `MongoClient.close()` does some nice things for us. Here's its JavaDoc:

```java
    /**
     * Close the client, which will close all underlying cached resources, including, for example,
     * sockets and background monitoring threads.
     */
    void close();
```

helloworld12321 (approved):
Looks great. Those are nice changes, @NicMcPhee! 🙂

server/src/main/java/umm3601/Server.java
floogulinc: Is there a reason you choose `serverStopping` rather than `serverStopped`?

Right now `server.stop()` is never called so that event hook never happens anyway. We can add this to actually shut down cleanly:
```java
Runtime.getRuntime().addShutdownHook(new Thread(() -> {
	app.stop();
}));
```

PR #320, opened 2021-02-19
Clean up the use of regexes for validation on the client-side

### LINKED ISSUE #313: Clean up the regexes in add-user.component.ts
cf. UMM-CSci-3601/3601-lab4-mongo#35.

Some of the regexes are a bit redundant (`^[0-9]+[0-9]?` for an age) and some of them are pretty weird. (`^[A-Za-z0-9\\s]+[A-Za-z0-9\\s]+$(\\.0-9+)?` for a human name). 🙃

Building off of @NicMcPhee's suggestions, I'd recommend

1. simplifying the age regex,
2. removing the user name regex (or, if you'd like, replacing it with something more unicode-friendly), and
3. adding a simple regex for email addresses.

(Maybe something like `[^@]+@[^@]+` for the email regex? I'd rather be overly permissive here than overly restrictive—we wouldn't want to accidentally forbid [plus extensions][], for example.)

[plus extensions]: https://helpspot.pobox.com/index.php?pg=kb.page&id=58

helloworld12321:
cf. #313 and UMM-CSci-3601/3601-lab4-mongo#35.

I've also changed a bit of how we're handling the age input, to make it clear what validation is coming from angular and what validation is coming from the HTML itself.

(TL;DR: We can rely on the HTML's built-in validation to check that the input is a number, but we have to add our own validation to check that the input is a _whole_ number.)

# Before

# After

Closes #313.

### DIFF
--- a/client/src/app/users/add-user.component.html
+++ b/client/src/app/users/add-user.component.html
@@ -23,10 +23,7 @@ is potentially helpful and contributed to the shaping of this example -->
 
           <mat-form-field>
             <mat-label>Age</mat-label>
-            <!-- We'd use type="number" here, but then the field would add some
-            of its own validation, which interferes with what we're doing. -->
-            <input matInput #input id="ageField" placeholder="Age"
-                  formControlName="age" required>
+            <input matInput #input type = "number" id="ageField" placeholder="Age"                  formControlName="age" required>
             <mat-error *ngFor="let validation of addUserValidationMessages.age">
               <mat-error class="error-message" id="age-error"
                         *ngIf="addUserForm.get('age').hasError(validation.type) && (addUserForm.get('age').dirty || addUserForm.get('age').touched)">
--- a/client/src/app/users/add-user.component.html
+++ b/client/src/app/users/add-user.component.html
@@ -23,7 +23,8 @@ is potentially helpful and contributed to the shaping of this example -->
 
           <mat-form-field>
             <mat-label>Age</mat-label>
-            <input matInput #input type = "number" id="ageField" placeholder="Age"                  formControlName="age" required>
+            <input matInput #input type="number" id="ageField" placeholder="Age"
+                  formControlName="age" required>
             <mat-error *ngFor="let validation of addUserValidationMessages.age">
               <mat-error class="error-message" id="age-error"
                         *ngIf="addUserForm.get('age').hasError(validation.type) && (addUserForm.get('age').dirty || addUserForm.get('age').touched)">
--- a/client/src/app/users/add-user.component.html
+++ b/client/src/app/users/add-user.component.html
@@ -23,7 +23,7 @@ is potentially helpful and contributed to the shaping of this example -->
 
           <mat-form-field>
             <mat-label>Age</mat-label>
-            <input matInput #input id="ageField" type="number" placeholder="Age"
+            <input matInput #input id="ageField" placeholder="Age"
                   formControlName="age" required>
             <mat-error *ngFor="let validation of addUserValidationMessages.age">
               <mat-error class="error-message" id="age-error"
--- a/client/src/app/users/add-user.component.html
+++ b/client/src/app/users/add-user.component.html
@@ -23,6 +23,8 @@ is potentially helpful and contributed to the shaping of this example -->
 
           <mat-form-field>
             <mat-label>Age</mat-label>
+            <!-- We'd use type="number" here, but then the field would add some
+            of its own validation, which interferes with what we're doing. -->
             <input matInput #input id="ageField" placeholder="Age"
                   formControlName="age" required>
             <mat-error *ngFor="let validation of addUserValidationMessages.age">
--- a/client/src/app/users/add-user.component.spec.ts
+++ b/client/src/app/users/add-user.component.spec.ts
@@ -166,7 +166,7 @@ describe('AddUserComponent', () => {
       expect(ageControl.hasError('max')).toBeTruthy();
     });
 
-    it('should not allow an age to contain non-digits', () => {
+    it('should not allow an age to be non-numeric', () => {
       // The HTML input field itself checks to make sure that the value is
       // numeric. Angular's form controls aren't even involved with this step.
       const ageHtmlInput =
@@ -174,6 +174,12 @@ describe('AddUserComponent', () => {
       ageHtmlInput.value = '27aa';
       expect(ageHtmlInput.validity.valid).toBeFalsy();
       expect(ageHtmlInput.validity.valueMissing).toBeTruthy();
+    });
+
+    it('should not allow an age to contain a decimal point', () => {
+      ageControl.setValue(27.5);
+      expect(ageControl.valid).toBeFalsy();
+      expect(ageControl.hasError('pattern')).toBeTruthy();
     });
   });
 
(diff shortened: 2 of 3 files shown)


client/src/app/users/add-user.component.html
```diff
@@ -23,7 +23,7 @@
 
           <mat-form-field>
             <mat-label>Age</mat-label>
-            <input matInput #input id="ageField" type="number" placeholder="Age"
+            <input matInput #input id="ageField" placeholder="Age"
```
floogulinc, line 26: I think it's pretty important to present a numerical input to the user. Can you elaborate on the issue you were seeing with this field having type "number"?

floogulinc: It probably makes sense to at least use `inputmode="numeric"` on the input like we do in Buzz About.

client/src/app/users/add-user.component.ts
```diff
@@ -72,7 +72,7 @@ export class AddUserComponent implements OnInit {
       // Since this is for a company, we need workers to be old enough to work, and probably not older than 200.
       age: new FormControl('', Validators.compose([
         Validators.required,
-        Validators.pattern('^[0-9]+[0-9]?'),
+        Validators.pattern('^[0-9]+$'),
```
floogulinc, line 75: I think the idea here was that they should be at least two digits but that is obviously captured by the min and max validation.

client/src/app/users/add-user.component.html
floogulinc: I would still prefer a real number input if possible because the browser's UI for it makes a lot of sense here.

client/src/app/users/add-user.component.ts
helloworld12321: Yeah, that seems like what they were going for :)

client/src/app/users/add-user.component.html
helloworld12321: That makes sense. If you use `type="number"` and then the user enters a non-numeric input, the form field will behave as though the user hasn't entered any value at all—it'll say "Age is required" instead of "Age must be a number".

The alternative, which I'd also be fine with, is to rely on the validation inherent to `input type="number"`, and not in add our own validation. 

(We'd have to do a bit of fiddling with the tests so that they understand the validation inherent to `input type="number"`, though—the fact that we'd have to refactor the tests a bit is part of the reason I avoided going that route.)

helloworld12321: Oh, also, the inherent validation messages are not as nice as the ones we can provide ourselves.

floogulinc: I think the fact the user simply can't enter non-numerical values in a number field is OK. Does the rest of the validation work, eg if a user enters 0 or -20?

helloworld12321: Yes, the rest of the validation works just fine.

Thanks for your feedback! I'll try to change that over to using the built-in `type="number"` validation.

helloworld12321:
Sorry, I had some `git` difficulties around this point.

NicMcPhee:
I don't remember (& I'm on my phone) - do we check that's it's a _non-negative_ whole number?

client/src/app/users/add-user.component.html
```diff
@@ -23,8 +23,7 @@
 
           <mat-form-field>
             <mat-label>Age</mat-label>
-            <input matInput #input id="ageField" type="number" placeholder="Age"
-                  formControlName="age" required>
+            <input matInput #input type = "number" id="ageField" placeholder="Age"                  formControlName="age" required>
```
NicMcPhee, line 26: Do we want the spaces in the `type` assignment?

client/src/app/users/add-user.component.spec.ts
```diff
@@ -132,14 +126,27 @@ describe('AddUserComponent', () => {
...
       ageControl.setValue('');
       expect(ageControl.valid).toBeFalsy();
     });
 
-    it('should be fine with "27"', () => {
-      ageControl.setValue('27');
-      expect(ageControl.valid).toBeTruthy();
+    describe('should be fine with "27"', () => {
+      // Both the form control and the HTML itself enforce certain constraints
+      // what we can put in the age field. The HTML makes sure that the value
```
NicMcPhee, line 136: "on what"

client/src/app/users/add-user.component.spec.ts
```diff
@@ -132,14 +126,27 @@ describe('AddUserComponent', () => {
...
+      // is numeric, and the form control makes sure that it's in the right
+      // range.
+      it('...the form control', () => {
+        ageControl.setValue('27');
+        expect(ageControl.valid).toBeTruthy();
+      });
+
+      it('...and the input field itself', () => {
```
NicMcPhee, line 144: Should/can we separate these two tests? Would it make sense to test the form field as a separate set of tests?

That probably shouldn't block this PR - maybe it's a new issue, or maybe it just doesn't make sense. 

Thoughts?

client/src/app/users/add-user.component.spec.ts
```diff
@@ -159,8 +166,18 @@ describe('AddUserComponent', () => {
       expect(ageControl.hasError('max')).toBeTruthy();
     });
 
-    it('should not allow an age to contain non-digits', () => {
-      ageControl.setValue('123x567');
+    it('should not allow an age to be non-numeric', () => {
+      // The HTML input field itself checks to make sure that the value is
+      // numeric. Angular's form controls aren't even involved with this step.
```
NicMcPhee, line 171: Maybe this is an argument for _not_ separating out the form controls?

NicMcPhee (commented):
We should probably fix the one spacing issue, but otherwise this is another significant improvement!

client/src/app/users/add-user.component.html
helloworld12321: Oops, nope, thanks! 😂

client/src/app/users/add-user.component.spec.ts
helloworld12321: Oh, sure—I could move the HTMLy tests into a separate describe block all by themselves, away from the form control tests. Would you prefer that?

client/src/app/users/add-user.component.spec.ts
helloworld12321: I don't think I quite understand what you're saying. Could you speak more to that point?

client/src/app/users/add-user.component.spec.ts
helloworld12321: Oops, thanks :) Good catch!

helloworld12321:
> I don't remember (& I'm on my phone) - do we check that's it's a non-negative whole number?

Yes, I think the `^[0-9]+$` regex catches that! 🙂 (And it caught that before I made this PR, too.)

(So previously, you got the message, "Age must be a number". Now, you get the message, "Age must be a whole number".)

client/src/app/users/add-user.component.spec.ts
helloworld12321: (I think I'm going to go ahead and separate them out!)

client/src/app/users/add-user.component.spec.ts
NicMcPhee: Ignore this. Things have apparently changed a fair bit since I left his comment and its not relevant anymore. 😀

NicMcPhee (approved):
This looks great to me – thanks!

floogulinc (approved):

helloworld12321:
Thank you—merging! :)

PR #321, opened 2021-02-20
Change the `POST /api/users/new` endpoint to just `POST /api/users`

### LINKED ISSUE #317: Change the `POST /api/users/new` endpoint to `POST /api/users`
Just a small thing, but it might be a good change to make if we're teaching students how to structure APIs.

Generally, in RESTful APIs, your routes are supposed to represent resources, so they're all named after nouns, not verbs. <sub>(Yes, I know that "new" is an adjective, not a verb. Close enough! 🙃)</sub>

In particular, when you want to create a new object in a collection, it's conventional to send a POST message directly to the collection: https://restfulapi.net/http-methods/#post

helloworld12321:
Both on the client-side and server-side. 

This way, URLs are more like resources instead of actions. (Following RESTful API conventions.  😴💤)

Closes #317.

### DIFF
--- a/client/src/app/users/user.service.spec.ts
+++ b/client/src/app/users/user.service.spec.ts
@@ -175,13 +175,13 @@ describe('User service: ', () => {
     expect(userService.filterUsers(testUsers, { name: userName, company: userCompany }).length).toBe(1);
   });
 
-  it('addUser() calls api/users/new', () => {
+  it('addUser() posts to api/users', () => {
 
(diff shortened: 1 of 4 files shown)


floogulinc (approved):

NicMcPhee (approved):
Looks great to me. Thanks - a nice small (but meaningful) improvement.

helloworld12321:
Thanks very much—merging! :)

PR #322, opened 2021-02-20
Refactor JSON in JUnit tests

helloworld12321:
Okay, this is a *very* minor issue, but the formatting of the JUnit tests used to be kind of intimidating. (There are still a lot of escaped characters, but hopefully it should be a bit easier to read now! ☺️)

### DIFF
--- a/server/src/test/java/umm3601/mongotest/MongoSpec.java
+++ b/server/src/test/java/umm3601/mongotest/MongoSpec.java
@@ -68,24 +68,24 @@ public class MongoSpec {
     userDocuments = db.getCollection("users");
     userDocuments.drop();
     List<Document> testUsers = new ArrayList<>();
-    testUsers.add(Document.parse("{\n" +
-      "                    name: \"Chris\",\n" +
-      "                    age: 25,\n" +
-      "                    company: \"UMM\",\n" +
-      "                    email: \"chris@this.that\"\n" +
-      "                }"));
-    testUsers.add(Document.parse("{\n" +
-      "                    name: \"Pat\",\n" +
-      "                    age: 37,\n" +
-      "                    company: \"IBM\",\n" +
-      "                    email: \"pat@something.com\"\n" +
-      "                }"));
-    testUsers.add(Document.parse("{\n" +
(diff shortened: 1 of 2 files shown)


floogulinc:
I noticed some of these don't have the keys in quotes so they aren't actually valid JSON. I guess the parser doesn't mind but maybe we should either make them valid JSON or switch to building the documents using `.append()` chains.

helloworld12321:
Oh my gosh, I didn't even notice! Let me go fix that.

floogulinc (approved):
Thanks

floogulinc (approved):

NicMcPhee (approved):
SOOOO much better! 😀

Anyone got any ideas on how to make the various `newTestUser`s at the end less intimidating?

floogulinc:
> SOOOO much better! 😀
> 
> Anyone got any ideas on how to make the various `newTestUser`s at the end less intimidating?

Since those tests are testing the creation of users I think we want to just have the JSON the "client" is sending like that.

helloworld12321:
> > SOOOO much better! 😀
> > Anyone got any ideas on how to make the various `newTestUser`s at the end less intimidating?
> 
> Since those tests are testing the creation of users I think we want to just have the JSON the "client" is sending like that.

Yeah, I considered making those `Document`s too, but it seemed like kind of a waste if we were just going to convert them back into strings ¯\\\_(ツ)\_/¯

helloworld12321:
Since this has two approvals, I think I'm going to go ahead and merge it. Thanks, everyone!

helloworld12321:
> > > SOOOO much better! 😀
> > > Anyone got any ideas on how to make the various `newTestUser`s at the end less intimidating?
> > 
> > 
> > Since those tests are testing the creation of users I think we want to just have the JSON the "client" is sending like that.
> 
> Yeah, I considered making those `Document`s too, but it seemed like kind of a waste if we were just going to convert them back into strings ¯\_(ツ)_/¯

It's a bit of a pity that multiline strings are only available starting in [Java 13](https://openjdk.java.net/jeps/355)! 🙂

PR #417, opened 2021-04-08
Remove old E2E styling not used with Cypress

floogulinc:


NicMcPhee (approved):
I don't know if you're done with this (you hadn't requested any reviews), but I'm fine with removing this. It was always kinda weird in the old Protractor land, and definitely means nothing useful here.

floogulinc:
This is it, I just noticed it while looking at some student code and realized I forgot to remove it when switching this repo to Cypress.

PR #598, opened 2021-12-15
Update to node 16 and lockfileVersion 2

### LINKED ISSUE #597: Move package-lock.json to lockfileVersion 2
See https://github.com/UMM-CSci-3601/3601-lab3_angular-javalin/issues/664

helloworld12321:
See https://github.com/UMM-CSci-3601/3601-lab3_angular-javalin/issues/664

Closes #597


NicMcPhee (approved):
This looks great – thanks!

At least on my computer at home (MacOS), all this breaks horribly with v17 of node (which is what I have installed at home). If I use `nvm` to switch to v16, then it all behaves much better. There are still a *bucket* of warnings, but at least everything compiles and runs, and all the tests pass.

helloworld12321:
Oops, I think node 16 may not actually be compatible with Angular 11.

See: https://github.com/angular/angular/blob/2938008a2005db209e3e3cae1925b1a7e152bbbb/package.json#L11
And this compatibility matrix (whose accuracy I can't vouch for): https://gist.github.com/LayZeeDK/c822cc812f75bb07b7c55d07ba2719b3

Because of that, I might recommend upgrading Angular before merging this PR—then, hopefully the EBADENGINE warnings will go away 🙂

PR #602, opened 2022-01-14
Add Checkstyle via Gradle

### LINKED ISSUE #274: Add automated Java code review

NicMcPhee:
This adds Checkstyle to the Gradle configuration so that programs won't build if there are any warnings from Checkstyle.

I used the "traditional" Sun configuration as the starting point, but disabled a few checks that experience suggests are more confusing than helpful.

I had to make a variety of (mostly small) changes to the existing code base so that we actually pass these checks now.

As a consequence of dealing with some warnings (in this case from the VS Code SonarLint plugin), I replaced all the uses of the Guava library with similar functionality that is now available in more recent versions of Java. I then removed Guava as a dependency since we don't need it anymore.

Closes #274


server/build.gradle
```diff
@@ -25,9 +28,6 @@ repositories {
 
 // External dependencies that our application utilizes
 dependencies {
-  // Google core libraries for Java, various useful utilities
```
NicMcPhee, line 28: We were only using the `ImmutableMap.of()` method from Guava, and that functionality is now provided by `java.util.Map.of()`. So I replaced all the uses of `ImmutableMap.of()` and removed this dependency.

server/config/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,225 @@
...
+  <!-- Checks for Headers                                -->
+  <!-- See https://checkstyle.org/config_header.html   -->
+  <!-- <module name="Header"> -->
+  <!--   <property name="headerFile" value="${checkstyle.header.file}"/> -->
+  <!--   <property name="fileExtensions" value="java"/> -->
+  <!-- </module> -->
+
+  <!-- Allows us to suppress warnings with @SupressWarnings annotations. -->
+  <!-- Requires the "SuppressWarningsHolder" module below in the "TreeWalker" section. -->
+  <module name="SuppressWarningsFilter" />
```
NicMcPhee, line 101: You need this and the `SuppressWarningsHolder` a few lines later in order to be able to use the `@SuppressWarnings({…})` annotation.

server/config/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,225 @@
...
+    <module name="InnerAssignment"/>
+    <module name="MagicNumber"/>
+    <module name="MissingSwitchDefault"/>
+    <module name="MultipleVariableDeclarations"/>
+    <module name="SimplifyBooleanExpression"/>
+    <module name="SimplifyBooleanReturn"/>
+
+    <!-- Checks for class design                         -->
+    <!-- See https://checkstyle.org/config_design.html -->
+    <!-- This only really makes sense if we're building a reasonably
```
NicMcPhee, line 186: The three rules I disabled are all here in the "class design" section.

server/src/main/java/umm3601/user/User.java
```diff
@@ -3,9 +3,29 @@
 import org.mongojack.Id;
 import org.mongojack.ObjectId;
 
+// There are two examples of suppressing CheckStyle
```
NicMcPhee, line 6: I tried to document when I was suppressing warnings so students would have a sense of when/why one might basically turn off the airbags.

server/src/main/java/umm3601/user/UserController.java
```diff
@@ -113,24 +113,59 @@ public void getUsers(Context ctx) {
...
+   * at a unique avatar image based on a user's email.
+   *
+   * This uses the service provided by gravatar.com; there
+   * are numerous other similar services that one could
+   * use if one wished.
+   *
+   * @param email the email to generate an avatar for
+   * @return a URI pointing to an avatar image
+   */
+  private String generateAvatar(String email) {
```
NicMcPhee, line 159: I extracted this method to try to clarify what was going on here.

server/src/test/java/umm3601/user/UserControllerSpec.java
```diff
@@ -118,35 +138,29 @@ public void setupEach() throws IOException {
...
 
-  @AfterAll
-  public static void teardown() {
-    db.drop();
-    mongoClient.close();
-  }
-
   @Test
-  public void GetAllUsers() throws IOException {
+  public void getAllUsers() throws IOException {
```
NicMcPhee, line 148: Almost every method name in this class started with an upper case letter, when only classes should start with upper case letters. That was kinda tedious to fix, actually.

server/src/test/java/umm3601/user/UserControllerSpec.java
```diff
@@ -37,43 +37,64 @@
...
-
-  private UserController userController;
-
-  private ObjectId samsId;
-
-  static MongoClient mongoClient;
-  static MongoDatabase db;
-
-  static ObjectMapper jsonMapper = new ObjectMapper();
+  // The client and database that will be used
```
NicMcPhee, line 48: I rearranged/restructred the fields and the setup/teardown code to try to get related pieces closer together. I don't know if that actually helped, but I found it easier to keep track of this way.

server/src/main/java/umm3601/user/UserController.java
```diff
@@ -113,24 +114,61 @@ public void getUsers(Context ctx) {
...
+
+    userCollection.insertOne(newUser);
+    // 201 is the HTTP code for when we successfully
+    // create a new resource (a user in this case).
+    // See, e.g., https://developer.mozilla.org/en-US/docs/Web/HTTP/Status
+    // for a description of the various response codes.
+    // The class HttpURLConnection contains named constants with
+    // most, but not all, of the standard HTTP status codes,
+    // including HTTP_CREATED for 201.
+    ctx.status(HttpURLConnection.HTTP_CREATED);
```
NicMcPhee, line 147: I switched to using named constants for the HTTP response codes. We should probably try to do something like this in the JS/TS code, but I haven't done the hunting to find a library that defines all those named constants.

NicMcPhee:
The previous commit adds config for a Checkstyle plugin for VS Code. It's not a great implementation, but it does appear to work. It requires a local install of the `checkstyle` executable, which will no doubt trip up folks setting things up on their own gear. Maybe I just don't bother with that? Thoughts?

floogulinc:
> The previous commit adds config for a Checkstyle plugin for VS Code. It's not a great implementation, but it does appear to work. It requires a local install of the `checkstyle` executable, which will no doubt trip up folks setting things up on their own gear. Maybe I just don't bother with that? Thoughts?

Have you looked at this other checkstyle extension? It has way more installs than the one you added and it looks like it automatically downloads checkstyle for you.

.vscode/extensions.json
```diff
@@ -36,6 +36,10 @@
     // https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker
     "streetsidesoftware.code-spell-checker",
 
+    // Checkstyle Java style checker
+    // https://marketplace.visualstudio.com/items?itemName=rianadon.vscode-java-checkstyle
+    "rianadon.vscode-java-checkstyle",
```
helloworld12321, line 41: (Cross-post from Slack)

> It requires that you have the checkstyle executable on your computer, which is likely to trip up folks trying to run things on their own computers. How much do we care about VS Code support?

Oof, that's no fun.

I installed the checkstyle executable from homebrew and looked at it—seems like it's just a tiny wrapper around the jar. :upside_down_face:

Unfortunately, there doesn't seem to be an equivalent "checkstyle executable" for Windows 🙁

Screen Shot 2022-01-14 at 12 59 58 PM

helloworld12321: We could always commit the checkstyle jar to our repo somewhere, and then point the `rianadon-checkstyle` extension at that jar. It's not a pretty solution, but it's simple, at least.

(This does mean that the version of checkstyle used in VS Code might drift from the version used in gradle. But that's also true if you installed checkstyle via Homebrew.)

helloworld12321: If you don't want to go down that route, we can always just not have a VS Code extension. (I guess that'll teach students how to read error messages, at least ¯\\\_(ツ)\_/¯)

server/config/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,225 @@
...
+    <module name="EmptyBlock"/>
+    <module name="LeftCurly"/>
+    <module name="NeedBraces"/>
+    <module name="RightCurly"/>
+
+    <!-- Checks for common coding problems               -->
+    <!-- See https://checkstyle.org/config_coding.html -->
+    <module name="EmptyStatement"/>
+    <module name="EqualsHashCode"/>
+    <module name="HiddenField"/>
```
helloworld12321, line 175: Would you be willing to set the

- `ignoreConstructorParameter`,
- `ignoreSetter`, and
- `setterCanReturnItsClass`

options to true?

I find it's often useful to write code like

```java
public Point(int x, int y) {
  this.x = x;
  this.y = y;
}
```

instead of having to choose new names, like

```
public Point(int xValue, int yValue) {
  x = xValue;
  y = yValue;
}
```

If you'd prefer to leave those options off, though, no worries! :)

server/config/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,225 @@
...
+    <!-- https://checkstyle.org/config_filters.html#SuppressionXpathFilter -->
+    <module name="SuppressionXpathFilter">
+      <property name="file" value="${org.checkstyle.sun.suppressionxpathfilter.config}"
+                default="checkstyle-xpath-suppressions.xml" />
+      <property name="optional" value="true"/>
+    </module>
+
+  </module>
+
+</module>
```
helloworld12321, line 225: I really like the set of rules you've chosen—these look good!

server/src/main/java/umm3601/user/UserController.java
```diff
@@ -113,24 +114,61 @@ public void getUsers(Context ctx) {
    * @param ctx a Javalin HTTP context
    */
   public void addNewUser(Context ctx) {
+    /*
+     * The follow chain of statements uses the Javalin validator system
+     * to verify that instance of `User` provided in this context is
+     * a "legal" user. It checks the following things (in order):
+     *    - The user has a value for the name (`usr.name != null`)
+     *    - The user name is not blank (`usr.name.length > 0`)
+     *    - The provided email is value (matches EMAIL_REGEX)
```
helloworld12321, line 123: Quick typo fix

```suggestion
     *    - The provided email is valid (matches EMAIL_REGEX)
```

server/config/checkstyle/checkstyle.xml
helloworld12321:
```suggestion
    <module name="HiddenField">
      <property name="ignoreConstructorParameter" value="true"/>
      <property name="ignoreSetter" value="true"/>
      <property name="setterCanReturnItsClass" value="true"/>
    </module>
```

server/src/main/java/umm3601/user/UserController.java
helloworld12321: Oh sweet, thanks! :)

server/src/test/java/umm3601/mongotest/MongoSpec.java
```diff
@@ -34,15 +46,21 @@
...
- * <p>
- * Created by mcphee on 20/2/17.
  */
+// The tests here include a ton of "magic numbers" (numeric constants).
+// It wasn't clear to me that giving all of them names would actually
+// help things. The fact that it wasn't obvious what to call some
+// of them says a lot. Maybe what this ultimately means is that
+// these tests can/should be restructured so the constants (there are
+// also a lot of "magic strings" that Checkstyle doesn't actually
+// flag as a problem) make more sense.
```
helloworld12321, line 56: I certainly have seen test classes that start by defining thirty different constants for sample data—but I agree with you that we're probably okay if we don't do that! 🙂

server/src/test/java/umm3601/user/UserControllerSpec.java
helloworld12321: Wahoo! Those names have been weird for a while—thanks for decapitalizing them.

helloworld12321 (changes requested):
Thanks so much for doing that. I'm glad to have a linter for the server-side!

This looks really good. There's one rule I wanted to ask about (about shadowing fields with parameters) and then the VS Code extension needs to get worked out, but other than that this looks ready to go! :)

helloworld12321:
Oh, and also it would be nice to mention how to invoke the checkstyle through gradle! 🙂

.vscode/extensions.json
NicMcPhee: I think I'm going to revert the VS Code part of this as I think that's just more complication than we really need. Thanks for looking into things and for the feedback.

server/config/checkstyle/checkstyle.xml
NicMcPhee: I wish I could take more credit, but these are really just the "old" Sun rules, which I've generally found quite useful.

README.md
```diff
@@ -46,6 +46,9 @@ Instructions on setting up the development environment and working with the code
 From the `server` directory:
 - `./gradlew run` to start the server
 - `./gradlew test` to test the server
+- `./gradlew checkstyleMain` to run Checkstyle on the server Java code in the `src/main` folder
```
NicMcPhee, line 49: I added examples of how to run Checkstyle on the CLI via Gradle. Thanks to @helloworld12321 for the suggestion.

NicMcPhee:
I think I've dealt with all the excellent suggestions from @helloworld12321. I also undid the changes that required and configured the VS Code Checkstyle extension since that seemed to be a bit of a mess.

helloworld12321 (approved):
Looks excellent. Thanks so much for your work on this!

NicMcPhee:
Thanks for all your help and feedback!

PR #609, opened 2022-01-15
Fix formatting of all-contributors table

floogulinc:
The previous manual merging had three rows in the table when its supposed to be all on one row. I used [`all-contributors-cli`](https://allcontributors.org/docs/en/cli/overview) to regenerate the README section. Specifically `npx all-contributors-cli generate`.


README.md
```diff
@@ -1,6 +1,6 @@
 !["Broken" badge to remind us to fix the URLs on the "real" badges](https://img.shields.io/badge/FIX_BADGES-Badges_below_need_to_be_updated-red)
 <!-- ALL-CONTRIBUTORS-BADGE:START - Do not remove or modify this section -->
-[![All Contributors](https://img.shields.io/badge/all_contributors-3-green.svg?style=flat-square)](#contributors-)
+[![All Contributors](https://img.shields.io/badge/all_contributors-3-orange.svg?style=flat-square)](#contributors-)
```
NicMcPhee, line 3: Do we know if they ever use a color other than orange here? I found that orange looked a little "warning – danger" to me, so I just switched to green. I don't really _mind_ orange, but in the context of our other uses of badges it does tend to look like a "bad" thing.

NicMcPhee (approved):
Great job on fixing this. Obviously my "by hand" repairs weren't perfect, and this is definitely nicer. I'd wondered why it was stacking all the boxes in a column when the documentation suggested that they should be in a row, and it turns out it was because _I'd_ stacked them. 😜

NicMcPhee (approved):

PR #610, opened 2022-01-15
Update many client-side dependencies at once

helloworld12321:
The update to lockfile-version 2 means that a lot of our dependabot pull requests have merge conflicts. So, I'm trying to update a whole bunch of client-side dependencies in one fell swoop.


NicMcPhee:
Thanks a ton for looking into this. I was just looking at something else for this, and ran into issues with `package-lock.json` that I think/hope this will take care of!

helloworld12321:
@NicMcPhee oh sure, but don't thank me yet, I still need to resolve a lot of issues :)

helloworld12321:
FYI:

Right now, we're on Jasmine 3.x.

Jasmine 4.x came out a week or two ago. It's stricter than Jasmine 3.x, in that it treats some async operations as errors (for example, trying to call a `done()` callback twice).

Jasmine `3.99.0` issues warnings in these cases.

Unfortunately, our project triggers some of these warnings. (I think one or more of our dependencies are responsible.)

I'm not sure if there's a way to fix these warnings. We may have to wait for our dependencies to fix the issue upstream.

In the meantime, I've pinned Jasmine to 3.10.x, which does not issue deprecation warnings.

helloworld12321:
Okay, I think that's the majority of the client-side updates! 🙂

@NicMcPhee, could I ask you to make sure this works locally? (I had trouble installing mongo, so I wasn't able to test it manually. I just checked that `npm install` worked and the CI passed.)

Thanks so much! :)

client/src/test.ts
```diff
@@ -9,8 +9,8 @@ import {
 
 declare const require: {
   context(path: string, deep?: boolean, filter?: RegExp): {
-    keys(): string[];
     <T>(id: string): T;
+    keys(): string[];
```
helloworld12321, line 13: The new version of ESLint asked us to reorder these declarations.

client/package.json
```diff
@@ -13,52 +13,52 @@
...
     "karma-chrome-launcher": "~3.1.0",
-    "karma-coverage-istanbul-reporter": "~3.0.2",
-    "karma-jasmine": "~4.0.0",
-    "karma-jasmine-html-reporter": "^1.5.4",
-    "ts-node": "~8.3.0",
-    "typescript": "~4.0.7"
+    "karma-coverage": "^2.1.0",
+    "karma-jasmine": "~4.0.1",
+    "karma-jasmine-html-reporter": "^1.7.0",
+    "ts-node": "^9.1.1",
```
helloworld12321, line 61: I would upgrade to ts-node 10.x.x, but that seemed to break our e2e tests.

(At least, I think it was ts-node that was causing the issue. It may have been an intermittent failure.)

client/karma.conf.js
```diff
@@ -9,21 +9,28 @@ module.exports = function (config) {
...
       require('karma-jasmine-html-reporter'),
-      require('karma-coverage-istanbul-reporter'),
+      require('karma-coverage'),
       require('@angular-devkit/build-angular/plugins/karma')
     ],
     client: {
       clearContext: false // leave Jasmine Spec Runner output visible in browser
     },
-    coverageIstanbulReporter: {
+    coverageReporter: {
```
NicMcPhee, line 18: Thanks for fixing this – this warning had definitely been catching my attention.

client/package.json
NicMcPhee: Do we want to create an issue to remind us to look into this later? Or is Dependabot already fussing about it?

client/package.json
```diff
@@ -13,52 +13,52 @@
...
+    "@angular-eslint/builder": "~12.7.0",
+    "@angular-eslint/eslint-plugin": "~12.7.0",
+    "@angular-eslint/eslint-plugin-template": "~12.7.0",
+    "@angular-eslint/schematics": "~12.7.0",
+    "@angular-eslint/template-parser": "~12.7.0",
+    "@angular/cli": "~12.2.15",
+    "@angular/compiler-cli": "~12.2.15",
+    "@angular/language-service": "~12.2.15",
     "@nrwl/cypress": "^12.6.3",
     "@nrwl/workspace": "^11.4.0",
```
NicMcPhee, line 43: I was slightly surprised this didn't get updated. This is the (indirect) source of [one of the security alerts](https://github.com/UMM-CSci-3601/3601-iteration-template/security/dependabot/client/package-lock.json/axios/open) that I had looked at for a second and was thinking it would get swept up in all this updating. No worries. We can deal with that later in another round of updates.

client/src/styles.scss
```diff
@@ -1,13 +1,12 @@
-
 // Custom Theming for Angular Material
+@use '~@angular/material' as mat;
```
NicMcPhee, line 2: Does the comment right above this still make sense here? There's no mention of theming in this import anymore, so I'm not sure.

client/src/test.ts
NicMcPhee: Interesting. Any idea why?

NicMcPhee (approved):
This looks great – thanks _sooo_ much!

I consider anything I added as questions rather than meaningful concerns. Feel free to merge away!

client/package.json
NicMcPhee: For the record, `@nrwl/workspace` is up into v13, and the version that fixes the linked security alert is somewhere in the v12 family.

client/src/test.ts
helloworld12321: Right, yes—the declaration at the top overloads the function call operator `()`

(It's the TypeScript equivalent of C++'s `operator()(string id)`.)

Since it's such a special declaration, ESLint wants it to have top billing.

client/package.json
helloworld12321: We certainly could. (I haven't looked into it enough to know what's up. 🤷🏻‍♂️)

I think dependabot will keep fussing about it, though.

client/package.json
helloworld12321: Oh, you're right! Yes, it would definitely make sense to update these two. (I may do that in a separate PR.) Thanks! 🙂

client/src/styles.scss
helloworld12321: Hrm, yes, it could do with being reworded, but it's not wrong, either. I'd be inclined to leave it.

Good catch, though!

PR #613, opened 2022-01-16
Update @nwrl/cypress and @nwrl/workspace

helloworld12321:
to the latest versions. This fixes some security issues in our transitive dependencies.


NicMcPhee (approved):
Thanks for doing this!

PR #646, opened 2022-01-21
Remove duplicate dependency

helloworld12321:
Oops, I accidentally resolved a merge conflict wrong and included two versions of the jackson-datatype-jsr310 package. This pull request removes the older one. Sorry about that! :(


NicMcPhee (approved):
Thanks for catching that!

PR #659, opened 2022-01-25
Upgrade to Javalin 4

NicMcPhee:
This makes all the necessary changes to move our code from Javalin 3 to 4, based largely on [the changes in Lab 4](https://github.com/UMM-CSci-3601/3601-lab4-mongo/pull/22).


server/src/main/java/umm3601/user/UserController.java
```diff
@@ -73,44 +75,57 @@ public void getUser(Context ctx) {
...
   /**
    * Get a JSON response with a list of all the users.
    *
    * @param ctx a Javalin HTTP context
    */
   public void getUsers(Context ctx) {
+    Bson combinedFilter = constructFilter(ctx);
+    Bson sortingOrder = constructSortingOrder(ctx);
+
+    // All three of the find, sort, and into steps happen "in parallel" inside the
```
helloworld12321, line 87: For my own benefit, could I ask about what "in parallel" means here? (Thanks! 😃)

server/src/test/java/umm3601/user/UserControllerSpec.java
```diff
@@ -152,74 +160,136 @@ public void setupEach() throws IOException {
...
 
-  @Test
-  public void getAllUsers() throws IOException {
+  /**
+   * Construct an instance of `ContextUtil`, which is essentially
+   * a mock context in Javalin. See `mockContext(String, Map)` for
+   * more details.
+   */
+  private Context mockContext(String path) {
+    return mockContext(path, Map.of());
```
helloworld12321, line 169: Consider using `Collections.emptyMap()` here (just for readability 🙂)

server/src/test/java/umm3601/user/UserControllerSpec.java
```diff
@@ -152,74 +160,136 @@ public void setupEach() throws IOException {
...
 
   /**
-   * Test that if the user sends a request with an illegal value in
-   * the age field (i.e., something that can't be parsed to a number)
-   * we get a reasonable error code back.
-   */
+  * Test that if the user sends a request with an illegal value in
+  * the age field (i.e., something that can't be parsed to a number)
+  * we get a reasonable error code back.
+  */
```
helloworld12321, line 270: Quick whitespace fix 🙂

```suggestion
   * Test that if the user sends a request with an illegal value in
   * the age field (i.e., something that can't be parsed to a number)
   * we get a reasonable error code back.
   */
```

server/src/test/java/umm3601/user/UserControllerSpec.java
```diff
@@ -260,25 +329,20 @@ public void getUsersByCompanyAndAge() throws IOException {
 
   @Test
   public void getUserWithExistentId() throws IOException {
-
     String testID = samsId.toHexString();
+    Context ctx = mockContext("api/users/:id", Map.of("id", testID));
```
helloworld12321, line 333: Just to check, should we also change the paths `"api/users/{id}"` in the test file?

server/src/test/java/umm3601/user/UserControllerSpec.java
```diff
@@ -365,13 +428,15 @@ public void addInvalidAgeUser() throws IOException {
...
 
-    assertThrows(BadRequestResponse.class, () -> {
+    Context ctx = mockContext("api/users");
+
+    assertThrows(ValidationException.class, () -> {
       userController.addNewUser(ctx);
     });
   }
 
+  @Test
```
helloworld12321, line 439: Lol :) nice catch!

helloworld12321 (changes requested):
Looks good! 🙂 Just a few questions and a few quick fixes.

Thanks so much for your work here!

server/src/test/java/umm3601/user/UserControllerSpec.java
```diff
@@ -152,74 +160,136 @@ public void setupEach() throws IOException {
     userController = new UserController(db);
   }
 
-  @Test
-  public void getAllUsers() throws IOException {
+  /**
+   * Construct an instance of `ContextUtil`, which is essentially
```
floogulinc, line 164: This isn't creating an instance of `ContextUtil` but rather an instance of `Context`.

server/src/test/java/umm3601/user/UserControllerSpec.java
```diff
@@ -152,74 +160,136 @@ public void setupEach() throws IOException {
...
+   */
+  private Context mockContext(String path) {
+    return mockContext(path, Map.of());
+  }
 
-    // Create our fake Javalin context
-    Context ctx = ContextUtil.init(mockReq, mockRes, "api/users");
-    userController.getUsers(ctx);
+  /**
+   * Construct an instance of `ContextUtil`, which is essentially a mock
```
floogulinc, line 173: This isn't creating an instance of `ContextUtil` but rather an instance of `Context`

server/src/main/java/umm3601/user/UserController.java
NicMcPhee: I wasn't thrilled about my use of "in parallel" (which was part of the use of the scare quotes, I suspect).

What I'm trying to get at is that MongoDB doesn't so the whole search, and _then_ sort the results, and _then_ put them in an array. Instead there's some streaming/lazy evaluation stuff going on. I don't don't if (a) that's particularly important or (b) I'm conveying that in a way that is remotely useful. Thoughts on better wording?

server/src/test/java/umm3601/user/UserControllerSpec.java
NicMcPhee: Totally! That would be much better – thanks!

server/src/test/java/umm3601/user/UserControllerSpec.java
NicMcPhee: Good catch! An example of updating code without updating the associated comments.

server/src/test/java/umm3601/user/UserControllerSpec.java
NicMcPhee: Yeah, there were _two_ of these "test" methods that didn't have the `@Test` annotation. Oops.

server/src/test/java/umm3601/user/UserControllerSpec.java
NicMcPhee: Oh, probably. As far as I can tell what you put here in the path is actually ignored unless your code uses it. So I'm not sure if we want `/{id}` or the actual ID or…

We definitely don't want `:id`, though.

server/src/test/java/umm3601/user/UserControllerSpec.java
floogulinc: This same comment is also wrong in Lab 4. `ContextUtil.init` is a static method that returns an instance of `Context`

floogulinc (approved):

helloworld12321 (approved):

PR #712, opened 2022-02-14
Upgrade to Angular 13 and Gradle 7.3.3

### LINKED ISSUE #603: Upgrade Gradle to v7.3.3
We presumably want to [upgrade this Gradle](https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:upgrading_wrapper) and all the other Java-based-labs to the latest version, which is currently 7.3.3.

We quite recently upgraded to 7.3.2, but of course they bumped up a minor version in late December…

### LINKED ISSUE #608: Update to Angular 13 and Angular-ESLint 13

### LINKED ISSUE #678: Migrate to official Cypress Angular schematic
There is now an official schematic for Cypress: https://www.npmjs.com/package/@cypress/schematic

We should migrate from `@nrwl/cypress` to `@cypress/schematic`. This will also allow us to remove the `@nrwl/workspace` dependency.

NicMcPhee:
Need to [upgrade from Angular 12 to 13](https://update.angular.io/?v=12.0-13.0). While I'm here I'll also:

* Switch the Cypress imports from `@nrwl/cypress` to the new `@cypress/schematics`
* Update Gradle to v7.3.3 so it matches the labs

Closes #603 
Closes #608 
Closes #678


client/package.json
```diff
@@ -13,41 +13,40 @@
...
+    "@angular-eslint/template-parser": "~13.1.0",
+    "@angular/cli": "~13.2.3",
+    "@angular/compiler-cli": "~13.2.2",
+    "@angular/language-service": "~13.2.2",
+    "@cypress/schematic": "^1.6.0",
     "@types/jasmine": "^3.8.2",
-    "@types/node": "^16.11.22",
-    "@typescript-eslint/eslint-plugin": "^5.10.2",
-    "@typescript-eslint/parser": "^5.12.0",
+    "@types/node": "^17.0.18",
```
helloworld12321, line 44: Could I ask you to downgrade `@types/node` to the latest 16.x.x version?

The `@types/node` 16.x.x line corresponds to Node 16, which is what we're using right now. 🙂

client/src/test.ts
```diff
@@ -17,7 +17,9 @@ declare const require: {
 // First, initialize the Angular testing environment.
 getTestBed().initTestEnvironment(
   BrowserDynamicTestingModule,
-  platformBrowserDynamicTesting()
+  platformBrowserDynamicTesting(), {
+    teardown: { destroyAfterEach: false }
+}
```
helloworld12321, line 22: Gosh, you'd think the auto-updater would be smarter about this 😜

```suggestion
  platformBrowserDynamicTesting(),
  {
    teardown: { destroyAfterEach: false }
  }
```

helloworld12321 (changes requested):
Looks great, just a few suggestions. 🙂

client/package.json
NicMcPhee: Good catch – I obviously got overexcited! 😜

helloworld12321 (dismissed):
LGTM, thanks!

client/angular.json
```diff
@@ -108,7 +108,7 @@
           }
         },
         "e2e": {
-          "builder": "@nrwl/cypress:cypress",
+          "builder": "@cypress/schematic:cypress",
           "options": {
```
floogulinc, line 112: The defaults when you add this schematic are to add these options:
```js
"watch": true,
"headless": false
```
This turns `ng e2e` from something that runs once headless to something that essentially runs `ng serve` and `cypress open` at the same time .

The behavior we are used to from `ng e2e` has been moved to the `ng run client:cypress-run`  from this new architect:
```json
"cypress-run": {
  "builder": "@cypress/schematic:cypress",
  "options": {
    "devServerTarget": "client:serve"
  },
  "configurations": {
    "production": {
      "devServerTarget": "client:serve:production"
    }
  }
},
```
It also by default adds a new `ng run client:cypress-open` command:
```json
"cypress-open": {
  "builder": "@cypress/schematic:cypress",
  "options": {
    "watch": true,
    "headless": false
  }
},
``` 

Though from my testing if we leave it as is everything still works as it did and that's probably fine. `ng e2e` runs the E2E tests once without opening the Cypress GIU (unexpectdly, setting only `"headless": false` in the options without `"watch": true` has no effect) and running `ng e2e --watch` runs the dev server in watch mode and opens the Cypress GUI.

So after all this, I actually don't think we should change anything here. You _could_ add the new commands but they're probably just redundant.

client/package.json
```diff
@@ -13,41 +13,40 @@
...
+    "@typescript-eslint/eslint-plugin": "5.12.0",
+    "@typescript-eslint/parser": "5.12.0",
     "cypress": "^9.4.1",
```
floogulinc, line 47: Interestingly when you add the new `@cypress/schematic` with `ng add` it sets this version to `latest` rather than any specific version.

floogulinc (changes requested):
I have some more changed I will need to do directly on the branch in a bit here.

NicMcPhee:
Somewhere in all of this (I think a Jasmine update) we now get several warnings of the form:

> ERROR: 'DEPRECATION: An asynchronous function called its 'done' callback more than once. This is a bug in the spec, beforeAll, beforeEach, afterAll, or afterEach function in question. This will be treated as an error in a future version. See<https://jasmine.github.io/tutorials/upgrading_to_Jasmine_4.0#deprecations-due-to-calling-done-multiple-times> for more information.
(in spec: User list has two users that are 37 years old)

@helloworld12321 indicated somewhere that apparently these warnings turn into flat out errors when we upgrade to v4 of Jasmine. Do we want to try to fix these problems here? Or push that off until later?

helloworld12321:
@NicMcPhee Hmm, interesting—what version of Jasmine are you running?

From what I know, the deprecation warnings should only appear on Jasmine 3.99.x and higher ([source][1]); we're pinning Jasmine to version 3.10.x to avoid this very issue (and this PR doesn't seem to change that 🤷🏻‍♂️).

[1]: https://github.com/jasmine/jasmine/blob/main/release_notes/3.99.0.md

helloworld12321:
@NicMcPhee Ah, okay, I get that too; it's coming from the version of Jasmine within `karma-jasmine`. I'm not sure quite what to do about that at the moment—I'll look into it! :slightly_smiling_face:

image

helloworld12321:
@NicMcPhee fixed the Jasmine warnings for now by running `npm dedupe`, which sets `jasmine-core` to the same version everywhere.

Screen Shot 2022-02-15 at 6 10 19 PM

Now, this does introduce a small peer-dependency issue.

Deduplicating npm packages sets `rxjs` to version 7.5.4 everywhere. However, one package (`rxjs-for-await`) wants to use `rxjs` version 6.x.x.

This doesn't seem to cause any problems; `rxjs-for-await` seems to work fine alongside `rxjs` 7.x.x. (And `rxjs-for-await` is only used by the linter, anyway.)

Screen Shot 2022-02-15 at 6 00 57 PM

helloworld12321 (approved):
With the Jasmine stuff taken care of—if only temporarily—this PR looks good to me! 🙂

I tested the new Docker settings, and they seem to work just fine. Thanks, @floogulinc!

helloworld12321 (approved):
Woo hoo!

floogulinc (approved):
I'm still not totally happy with the Jasmine situation but I think @helloworld12321's solution here is the way to go for this PR. Also thanks @helloworld12321 for testing the Docker stuff.

NicMcPhee:
We can definitely continue to work on improving the Jasmine tests, but I do think that should be in another PR than this one. I'm going to go ahead and merge this in with huge thanks to the two of you for all your help here!

PR #713, opened 2022-02-16
Upgrade to Jasmine 4

### LINKED ISSUE #612: Update to Jasmine 4
To update Jasmine, we may need to update some karma-related dependencies. See: https://github.com/UMM-CSci-3601/3601-iteration-template/pull/610#issuecomment-1013784234

helloworld12321:
Note that Angular's `waitForAsync` function doesn't play nice with Jasmine 4—see <https://github.com/UMM-CSci-3601/3601-iteration-template/pull/610#issuecomment-1013784234>—so we had to write a wrapper for `waitForAsync`, to work around the problematic behavior.

Closes #612.


helloworld12321:
Okay, there's a fix coming upstream (<https://github.com/angular/angular/pull/45025>) so we may not need to resort to this sort of kludge. 🙂

Thanks to @floogulinc for pointing this out!

NicMcPhee:
It seems like it would be useful to see if we can just pull in a newer version of `zone.js` and make this all go away. Also, no big surprise, but merging in the big Angular 13 PR (#712) created merge conflicts here @helloworld12321.

NicMcPhee:
As I said on Slack, maybe we just let this go for now? We're clearly not taking any advantage of the v4 features, and it's causing us at least a little grief, so maybe we just step away for now?

helloworld12321:
Closed; we're going to wait for the fix upstream ([in zone.js](https://github.com/angular/angular/pull/45025)) instead.

PR #794, opened 2022-03-14
Update client deps 2022-03-14

NicMcPhee:
This is an experiment in merging multiple Dependabot PRs in one PR to speed things up some on the review, merge, update, review, merge, update, etc., cycle. I was sort of hoping that I could just merge a bunch of Dependabot branches into one branch, but after a while you start to pile up merge conflicts. Those are easy enough to resolve in `package.json`, but they're a pain in `package-lock.json`. I might just start updating `package.json` by hand and let `npm install` update `package.json` appropriately.


NicMcPhee:
I realized that I (at least in principle) should have been using `ng update` when possible, so I scratched this and started over with #795.

PR #795, opened 2022-03-14
Update client dependencies 2022-03-14

NicMcPhee:
This updates all the client dependencies except for two that break things in different ways.

I used `ng update` wherever I could, and `npm update` for most of the rest.

The two packages that are still behind are `typescript` and `jasmine-core`. Upgrading `typescript straight up breaks things and the project won't compile (I haven't looked up why that might be), and upgrading `jasmine-core` leads to a bunch of warnings that we want to clean up before we do that (see #612).

This will presumably close a whole pile of Dependabot PRs.


helloworld12321 (approved):
Looks good to me!

I agree that our TypeScript version is good for now. A quick glance shows that Angular 13.2.6 uses TypeScript 4.5.x internally, which matches the TypeScript version we're using.

See: <https://github.com/angular/angular/blob/6f1c941dfec8cf0ea0ccbb0fda76f3d960fd843b/package.json#L156>

PR #874, opened 2022-04-22
Replace our use of JS string `+` with template literal

### LINKED ISSUE #836: Use JavaScript/TypeScript template strings
From @jpwalbran on Slack:

> This is a PSA about using typescript template strings, which are things that we haven’t been using and totally should. They allow for having custom formatted strings with values attached that are cleaner than using
> ```
> myString = 'string literal' + value + 'another string literal';
> ```
> Instead, you can use a template string to turn that same expression into
> ```
> myString = `string literal ${value} another string literal`;
> ```
> More information on them can be found [here](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html). 

We totally should be using/modelling this here in the template and in the labs.

NicMcPhee:
I'm just using VS Code's global search tool to look for +s in JS/TS code, and then convert them to use template literals.

The Karma tests didn't fail if I broke many of these, but the E2E tests did, which was nice since I forgot to switch to backticks at least once and actually broke things. The tests caught it, though, and I think this is good now.

Closes #836

### DIFF
--- a/client/src/app/users/add-user.component.ts
+++ b/client/src/app/users/add-user.component.ts
@@ -107,7 +107,7 @@ export class AddUserComponent implements OnInit {
   submitForm() {
     this.userService.addUser(this.addUserForm.value).subscribe(newID => {
       this.snackBar.open(
-        `Added User ${this.addUserForm.value.name}`,
+        `Added user ${this.addUserForm.value.name}`,
         null,
         { duration: 2000 }
       );
--- a/client/src/app/users/add-user.component.ts
+++ b/client/src/app/users/add-user.component.ts
@@ -106,9 +106,11 @@ export class AddUserComponent implements OnInit {
 
   submitForm() {
     this.userService.addUser(this.addUserForm.value).subscribe(newID => {
-      this.snackBar.open('Added User ' + this.addUserForm.value.name, null, {
-        duration: 2000,
-      });
+      this.snackBar.open(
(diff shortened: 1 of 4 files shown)


client/src/app/users/add-user.component.ts
```diff
@@ -106,9 +106,11 @@ export class AddUserComponent implements OnInit {
 
   submitForm() {
     this.userService.addUser(this.addUserForm.value).subscribe(newID => {
-      this.snackBar.open('Added User ' + this.addUserForm.value.name, null, {
-        duration: 2000,
-      });
+      this.snackBar.open(
+        `Added User ${this.addUserForm.value.name}`,
```
helloworld12321, line 110: Thoughts on lowercase here?

```suggestion
        `Added user ${this.addUserForm.value.name}`,
```

helloworld12321 (changes requested):
I'd suggest one letter case change, but otherwise looks great, thanks! 🙂

helloworld12321 (approved):
LGTM! (Although you may need to re-run Cypress)

NicMcPhee:
Poop – I'm pretty sure this actually breaks the test, which I'm guessing has that text (with the "wrong" case) hardwired in. Doh!

I'll fix that and equest a re-review.

helloworld12321 (approved):

PR #927, opened 2022-05-14
Update Java libraries

NicMcPhee:
This handles the three outstanding Dependabot updates for Java libraries. Since none of those Dependabot PRs ever got checked (because BetterCodeHub was having a bad day when they were created), I figured it would be easier to just do them in on new PR than force "fake" commits to each of the Dependabot PRs.

This updates:

* Javalin from 4.4.0 to 4.5.0
* mongodb-driver-sync from 4.5.1 to 4.6.0
* mongojack from 4.3.0 to 4.5.0

This should clear out three Dependabot PRs (#906, #907, and #908).

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -29,16 +29,16 @@ repositories {
 // External dependencies that our application utilizes
 dependencies {
   // Javalin, a simple web framework for Java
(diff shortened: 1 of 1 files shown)


helloworld12321 (approved):
LGTM, thanks! 🙂

PR #949, opened 2022-05-20
Client dependency updates, May 2022

NicMcPhee:
This uses `ng update` to updated a number of Angular libraries, and then `npm` and `ncu` to udate a whole pile of other client-side dependencies.

I _think_ I managed to skip updating Jasmine to 4.x and Node past v16, but someone should probably confirm that.

All the tests pass, so hopefully this is good.


helloworld12321 (approved):
Looks good, thanks so much!

PR #976, opened 2022-06-04
Update to Angular 14

### LINKED ISSUE #974: Upgrade to Angular 14
[Angular 14 is out!](https://blog.angular.io/angular-v14-is-now-available-391a6db736af) It doesn't look like [this upgrade](https://update.angular.io/?l=2&v=13.0-14.0) will break or change much more us and should be pretty straightforward.

There are some things we'll probably want to think about or do something about/with, though, including:

- Do we want to (eventually) do anything with the new standalone component option? This looks like an effort to make it easier to spin up something simple (which things like React and Due tend to be better for), but I don't immediately think that matters for Software Design? Maybe we could start with something simpler using the standalone components and then expand to a "full" Angular app after that? It's only in Preview mode now, though, so we definitely don't do anything until it stabilizes (if even then).
- We'll want to convert to Typed Angular Forms. It looks like the update process will convert any existing form thingies into _untyped_ form thingies, and we'll then want to convert those untyped thingies into typed thingies.
- They have [new change detection guides](http://angular.io/guide/change-detection), and we might want to incorporate those into the write-ups in some way. We've never really talked about change detection much (if at all), but it could help demystify some of the Angular features and improve performance somewhat.
- We should use the new page title features in the router in our examples and encourage students to do the same, as this helps with accessibility.
- There's a new ["Accessibility in Angular"](https://www.youtube.com/watch?t=431&v=Gm_bD1abFXI&feature=emb_imp_woyt) video from Google I/O 2022 that might be a useful resource to include.
- Say something/more about the Angular DevTools in the write-ups (here and earlier). This extension is now available for both Chrome and Firefox, so we really ought to talk about it and demonstrate its usage.

NicMcPhee:
This upgrades this project to Angular 14.

I did this by running (as instructed by [the upgrade guide](https://update.angular.io/?l=2&v=13.0-14.0)):

```text
ng update -C @angular/core@14 @angular/cli@14 --force
ng update -C @angular/material@14 --force
```

The `-C` flags tell `ng update` to automatically create commit messages at various stages in the process, which is very nice. The first `ng update` for example, creates 4 separate commits, one of which is specifically the changes to the untyped form elements, which makes it trivial to see exactly what we then need to address there.

I always seem to need the `--force` flags for these updates; the system _never_ seems to have a consistent set of libraries and dependencies. That's frankly kind of annoying, and if anyone can actually figure out how we can fix this that would be great.

While we only explicitly list 3 `@angular/*` packages in these updates, they end up dragging over a dozen other packages up with them. Running `npm outdated` after these makes it clear that there are no further `@angular` packages that need to be updated.

There's at least one spin-off change we'll have to deal with, but which I want to do in a separate issue/PR, namely converting the (now) untyped form elements to the new typed form elements. Commit 1be195dcb3c4b15646132babba7e53214958c180 explicitly shows which forms were converted to untyped forms in the update, and these are the things that will need to be converted to typed form elements.

Closes #974


helloworld12321 (approved):
LGTM! Sorry for taking a while to review this.

I'm sure you know the following, but just to say it out loud: we probably want to update `@angular-eslint/*` to version 14.x.x when that's released. (At the time of writing, `@angular-eslint/*` 14.x.x is still in alpha.)

As it stands, though, the linter still succeeds even with the `@angular-eslint/*` 13.x.x packages installed, so it looks like things are okay right now. 🙂

NicMcPhee:
Thanks for approving this! We definitely will have to update the linting when those packages are available. One advantage of Dependabot is that it'll let us know when the packages are ready.

PR #1007, opened 2022-06-17
Additional Angular 14 updates

NicMcPhee:
Various other packages (`@angular` and others) are updating to go along with the update to Angular 14. This tries to bring a lot of those together into one PR.

Unfortunately there's an inconsistency introduced by either #984 or #991 that means that `npm install` now fails everywhere. I could roll something back to make things work, but since no one's actually working on this at the moment my inclination is just to wait for the various dependent libraries to catch up and fix things. I'll keep this as a draft PR for now, and only switch it when we seem to have the necessary updates to get the whole thing to build properly.

I think at least part of the problem is that `@angular-eslint/schematics@13.5.0` doesn't like Angular v14. Hopefully they'll get `@angular-eslint/schematics` up to v14 soon and that will fix this. There may also be a problem because `@angular/flex-layout` isn't yet at v14.


NicMcPhee:
This gets everything updated to Angular v14 and the system finally builds and the tests run and pass.

Unfortunately BetterCodeHub now is grumpy because some of the test files have really long functions, so I need to do something about that.

client/package.json
```diff
@@ -13,51 +13,51 @@
...
     "@floogulinc/cypress-mongo-seeder": "^1.1.1",
-    "@types/jasmine": "^3.8.2",
-    "@types/node": "^16.11.26",
-    "@typescript-eslint/eslint-plugin": "^5.25.0",
-    "@typescript-eslint/parser": "^5.27.1",
-    "cypress": "^9.7.0",
-    "eslint": "^8.17.0",
+    "@types/jasmine": "^4.0.3",
+    "@types/node": "^18.0.0",
+    "@typescript-eslint/eslint-plugin": "^5.29.0",
```
floogulinc, line 46: Do we want to update the Node types to greater than the current LTS release we are using everywhere (16)?

client/package.json
```diff
@@ -13,51 +13,51 @@
...
+    "@cypress/schematic": "^2.0.0",
     "@floogulinc/cypress-mongo-seeder": "^1.1.1",
-    "@types/jasmine": "^3.8.2",
-    "@types/node": "^16.11.26",
-    "@typescript-eslint/eslint-plugin": "^5.25.0",
-    "@typescript-eslint/parser": "^5.27.1",
-    "cypress": "^9.7.0",
-    "eslint": "^8.17.0",
+    "@types/jasmine": "^4.0.3",
+    "@types/node": "^18.0.0",
```
floogulinc, line 45: Do we want to update the Node types to greater than the current LTS release we are using everywhere (16)?

floogulinc:
I'm not sure how Cypress fixtures would help with the code quality stuff. It might be reasonable to just exclude test files from those checks anyways since repetition and long functions aren't really bad in tests at least imo.

NicMcPhee:
> I'm not sure how Cypress fixtures would help with the code quality stuff. It might be reasonable to just exclude test files from those checks anyways since repetition and long functions aren't really bad in tests at least imo.

That's a fair point, and I did think about just excluding those from the checks. I'm open to potentially going down that road, but it wouldn't be my first choice. I think that our test code has gotten fairly messy over time and this might be a good excuse to do some housekeeping.

NicMcPhee:
> > I'm not sure how Cypress fixtures would help with the code quality stuff. It might be reasonable to just exclude test files from those checks anyways since repetition and long functions aren't really bad in tests at least imo.
> 
> That's a fair point, and I did think about just excluding those from the checks. I'm open to potentially going down that road, but it wouldn't be my first choice. I think that our test code has gotten fairly messy over time and this might be a good excuse to do some housekeeping.

I did a whole pile of refactoring to try to make BetterCodeHub happier, but it's still grumpy, and I'm no sure I'm actually making anything more readable, TBH. I'm leaning towards just excluding the E2E specs from BCH and undoing the refactoring that I did. I'm going to switch this back to a draft PR for now, and I'll try to tidy things up sometime tomorrow.

client/package.json
NicMcPhee: I don't know. Is the types system smart enough to know which version of Node that we're using and actually apply that set of types? I'm not sure how backwards compatible these tools are.

We presumably want to switch to Node v18 sometime anyway, so if this isn't breaking things we could just leave it.

floogulinc: It's always going to be safer to use types from older versions. Additionally, 18 is not a LTS release so we won't be switching to it. 16 is the current LTS release until 20 releases in April 2023.

floogulinc:
Yes I don't think the things BCH and other tools check for really apply to test code.

client/cypress/e2e/add-user.cy.ts
```diff
@@ -1,17 +1,9 @@
...
-  beforeEach(() => {
-    page.navigateTo();
-  });
-
-  it('Should have the correct title', () => {
-    page.getTitle().should('have.text', 'New User');
-  });
+const page = new AddUserPage();
 
+const checkAddUserButtonState = () => {
```
helloworld12321, line 6: @NicMcPhee I'm with @floogulinc, I'm afraid—this refactoring looks odd to me. 🤷🏻‍♂️

I would never expect a named function to declare a test case inside of it. That makes me have to jump around the source file to see what's actually being tested.

The problem is, BCH doesn't understand that Cypress, Jasmine, etc use a DSL. 🫤 A `describe(() => {})` block in Cypress is not the same conceptual object as a function, and there is no universe in which they should be held to the same length limits.

I don't know—does that make sense? 🤷🏻‍♂️

helloworld12321 (changes requested):

client/package.json
NicMcPhee: Thanks for the info about 18 not being LTS. I think I somehow assumed all the even numbered versions were LTS. Oops.

helloworld12321: > Thanks for the info about 18 not being LTS. I think I somehow assumed all the even numbered versions were LTS.

Oh, TIL, thanks! 🙂

NicMcPhee: So I've done some reading and I _think_ that we should be safe upgrading `@types/node` (or any of the `@types/*` dependencies) to higher versions than the thing that we're actually using. The typing system supports/expects information on when a feature was introduced/modified/etc., and seems to do the typing based on the major semver number. (I found [this](https://stackoverflow.com/questions/42035263/relationship-between-the-version-of-node-js-and-the-version-of-types-node) and [that](https://github.com/definitelytyped/definitelytyped#how-do-definitely-typed-package-versions-relate-to-versions-of-the-corresponding-library) useful, as well as just looking at some of the `*.d.ts` files [in `types/node`](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/node).)

So I think we should be fine having a newer version of a `@types` file, as it should (dangerous words, I realize) be able to recognize the versions we're actually using and apply the correct typing.

Dependabot "wants" us to upgrade `@types/node` (see #1037), so actually doing so will avoid an ongoing wrestling match with that, which would be nice.

So unless we have some evidence that the typing system doesn't use versions the way I think they are (or that that whole system is busted in some way), I'd be inclined to go ahead and let this version change happen.

Thoughts? (esp. @floogulinc)

client/cypress/e2e/add-user.cy.ts
NicMcPhee: It makes total sense, and I shared your concerns as I was doing the refactoring. I _like_ the `describe()/it()` structure that we get with Jasmine, and extracting those functions really didn't seem to improve readability for all the reasons you gave.

I'll unwind all that refactoring, and then we can see how we feel about things.

NicMcPhee:
So I think that the reason that BCH started being grumpy about the Cypress tests was the changes in the directory/naming schemes in moving to Cypress 10. It looks like `.bettercodehub.yml` had entries that were (I think) ignoring the test code before, and those needed to be updated when Cypress changed the naming system. This would make a lot of sense, as it wasn't at all clear to me why BCH would be grumpy _now_ when it had been happy enough before.

That said, I can't seem to get it to ignore those files. I've tried a number of changes, and it still seems to want to fuss about the E2E files. Hmph.

client/package.json
floogulinc: Typescript has no way to "detect" which version of node you're actually using. The types packages are aligned to the versions of the packages they are types for. Eh @types/node@18.x.x is types for Node.js 18 and @types/node@16.x.x is types for Node.js 16. There is no mechanism for typescript to know to use anything else. GitHub actually recently fixed dependabot to match your types with the actual version of the package you use: https://github.blog/changelog/2022-05-10-dependabot-keeps-types-dependencies-in-sync-with-updated-packages/
This doesn't work for Node itself though because Node itself isn't a npm package.

The problem with using newer types than the version of node we use for builds and testing is that users may attempt to use a feature only added in newer versions of Node and typescript will be happy because the types match what you've declared as the node version for the types.

floogulinc: Additionally, https://github.com/definitelytyped/definitelytyped#how-do-definitely-typed-package-versions-relate-to-versions-of-the-corresponding-library is talking about their internal publishing system and how it determines which version number to publish from a given set of types.

floogulinc: It seems GitHub is currently recommending you have dependabot ignore the @types/node package https://github.com/github-community/community/discussions/16339#discussioncomment-2917404

NicMcPhee: You've convinced me 😜. I've rolled back `@types/node` to v16 and added the bit telling Dependabot to ignore `@types/node` updates. That does mean that we'll have to remember to move that forward "by hand" when we do move to a new version of node, and that's exactly the kind of thing that I tend to forget, but we'll live.

client/cypress/e2e/add-user.cy.ts
helloworld12321: Roger that, thanks! 🙂

helloworld12321 (approved):
LGTM, thanks for doing that! 🙂

client/.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
+{
+    "cSpell.words": [
+        "DATAGENE",
+        "KINETICUT",
+        "MOMENTIA",
+        "OHMNET"
+    ]
+}
```
floogulinc, line 8: This file seems to have been created erroneously. There shouldn't be a separate vscode settings file just within client.

client/package.json
```diff
@@ -13,51 +13,51 @@
...
+    "@cypress/schematic": "^2.0.0",
     "@floogulinc/cypress-mongo-seeder": "^1.1.1",
-    "@types/jasmine": "^3.8.2",
-    "@types/node": "^16.11.26",
-    "@typescript-eslint/eslint-plugin": "^5.25.0",
-    "@typescript-eslint/parser": "^5.27.1",
-    "cypress": "^9.7.0",
-    "eslint": "^8.17.0",
+    "@types/jasmine": "^4.0.3",
+    "@types/node": "^16.0.0",
```
floogulinc, line 45: This probably should still be the latest minor version.

client/.vscode/settings.json
NicMcPhee: Good point – I'll clean that up.

client/package.json
NicMcPhee: Which "this"? I assume you mean that `@types/node` should be 16.x.y for the highest values of x and y? Agreed. I'll do that quick as well.

Issue #1047, opened 2022-07-10
Replace BetterCodeHub


NicMcPhee:
While I quite like BetterCodeHub in many ways, we've had a variety of issues that have convinced me that we probably want to look for other alternatives.

### Desired properties

* When possible, it would be nice to have checks that can be done in the IDE so that students get immediate feedback and don't have to wait for something to happen on GitHub (or elsewhere) to find out that there's an issue. This may not be possible everywhere (e.g., code duplication), but it would be nice when possible.
* Related to the previous, it would be good if things are fast; having to wait a zillion years for your GitHub Actions checks to complete is frustrating for everyone involved.
* It would be nice when possible to have the same tool handle both the client (JS/TS) code and the server (Java) code. That may not be very reasonable, though.

### Desired checks

BetterCodeHub does 10 checks, but we don't really care about all of them (at least not in this context). The ones that I'd like us to have checks for include:

* **Write short units of code**: This is essentially a simple length check. I know that this irks @floogulinc, but I think that for beginning coders (which Software Design students really are), having something that squawks when functions/classes start to get long is a Good Thing. In a perfect world it would be nice if the check supported something like "Beyond X is a warning, but beyond Y is an error". That way they could go a little over the line and things would still build, but they'd get poked about it and know that they need to be watching that component.
* **Write simple units of code**: I think this is basically a cyclomatic complexity check. Again, it would be nice to have both warning and error settings.
* **Write code once**: This is one of the most important checks IMO, and to be honest the students rarely noticed when BCH started warning them about issues there, so it would be nice to surface this more explicitly. This kind of analysis is harder to do on-the-fly, so I'm not sure if we'll find something that will work in real-time in the IDE, but it would be cool if we could.
* **Keep unit interfaces small**: This is basically about keeping the number of arguments in a function under control, and can probably be checked by the same thing that checks unit length and complexity. I'm generally a fan of this, although the way Angular uses injection in constructors frequently leads to quite a few arguments. It would be nice to have something that can be configured to ignore those constructors, but that might require too much Angular knowledge to be reasonable? And, to be fair, I would consider this check a fairly low priority; I don't see a lot of code that violates this in egregious or problematic ways, probably because of all the structure that Angular and Javalin provide.
* **Write clean code**: I think that this should be mostly handled by other linters and the LGTM checks, so it shouldn't be a major issue.

Some that I think we can ignore:

[...]

PR #1048, opened 2022-07-11
Replace BetterCodeHub

NicMcPhee:
We're experimenting with alternatives to BetterCodeHub (see #1047).


kklamberty:
Deep Source seems like it might be a decent replacement for BCH. It might be worth looking at. It is part of the GitHub Student Developer Pack and can be used on unlimited public repos (as well as 1 private repo).

NicMcPhee:
I actually think that DeepSource is more a replacement for LGTM than for BCH. LGTM and DeepSource both seem to be about more statement-level recommendations, where BCH was looking at more "big picture" design things like copy/pasted code, module dependencies, and the like.

That said, I think the experiments with DeepSource have been interesting and worth pursuing.

NicMcPhee:
CodeQL might be relevant here (see #1242), and the addition of Deep Source (#1299) may actually resolve this? I'd love to hear from the students on what they think might be most helpful.

PR #1109, opened 2022-08-09
Client dependency updates August, 2022

NicMcPhee:
This updates a number of Angular libraries to (mostly) v14.1.1, and updates other Node packages to their most recent version. This leaves `@types/node` at v16.11.47 since we're keeping the lab at Node v16 since that's the most recent LTS version.


helloworld12321 (approved):
LGTM, thanks! 😃

PR #1145, opened 2022-08-22
Client updates, late August 2022

NicMcPhee:
This updates various Angular and other NPM packages to the latest versions. No big version changes here, and I remembered to leave @types/node at v16. 😜


kklamberty:
> I get the same message even if I'm running Node v16, so it seems that there's some deep dependency that is doing something that is not (or in the future will not be) considered Good Form.
> 
> This seems to be [a known problem](https://stackoverflow.com/questions/73428935/errors-running-ng-lint-on-newly-created-angular-workspace), and the fix is ready for release in the next version of `@nrwl/devkit`. I think we just wait a bit and see if that update comes through soon.
> 
> Feel free to review if you wish, but we'll need a quick re-review after we make that change once it's available. Thanks!

The error and its fix is one that might be an interesting to share in the Foundations of Computer Science classes since it appears to just be a misplacement of parentheses with the not operator...
https://github.com/nrwl/nx/commit/6fdc123537b1fbf60657fee562dd1ff5f8863f4d

floogulinc:
I wonder what we have that's requiring `@nrwl/devkit` since we're not using any of their tools directly anymore.

NicMcPhee:
> I wonder what we have that's requiring `@nrwl/devkit` since we're not using any of their tools directly anymore.

I assume there's some way to ask `npm` to explain why we're downloading something?

NicMcPhee:
So it turns out that `@angular-eslint/builder` is what's pulling in `@nrwl/devkit`:

```text
❯ npm list @nrwl/devkit
client@0.0.0 /Users/mcphee/Documents/Courses/CSci3601/3601-iteration-template/client
└─┬ @angular-eslint/builder@14.0.3
  └── @nrwl/devkit@14.5.10
```

Folks obviously really jumped on this issue 🎉 and there are already updates that fix the `ng lint` issue.

Everything seems good in case someone has a second to provide a review so we can merge this in. Thanks!

floogulinc (approved):

PR #1167, opened 2022-08-31
Add ng-update action

NicMcPhee:
@floogulinc suggested that we try [the `ngxUptodate` action](https://github.com/fast-facts/ng-update) again. Hopefully this will run on the first day of every month, and create a PR that updates all the Angular libraries that are ready for updates.

If this works, it might make sense to run this more often (weekly?) since we'll be able to just approve the PRs that it generates and not hassle everyone with the need for code reviews.

### DIFF
--- a/.github/workflows/ng-update.yml
+++ b/.github/workflows/ng-update.yml
@@ -1,4 +1,4 @@
-name: "Update Angular Action"
+name: "Update Angular"
 on: # when the action should run. Can also be a CRON or in response to external events. see https://git.io/JeBz1
   schedule:
     - cron: '0 5 1 * *'
--- a/.github/workflows/ng-update.yml
+++ b/.github/workflows/ng-update.yml
@@ -2,6 +2,7 @@ name: "Update Angular Action"
 on: # when the action should run. Can also be a CRON or in response to external events. see https://git.io/JeBz1
   schedule:
     - cron: '0 5 1 * *'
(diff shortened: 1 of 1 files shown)


.github/workflows/ng-update.yml
```diff
@@ -0,0 +1,13 @@
+name: "Update Angular Action"
+on: # when the action should run. Can also be a CRON or in response to external events. see https://git.io/JeBz1
+  schedule:
+    - cron: '0 5 1 * *'
```
floogulinc, line 4:
```suggestion
    - cron: '0 5 1 * *'
  workflow_dispatch:
```

If you add this we can manually trigger the workflow without waiting for the right time to roll around.

.github/workflows/ng-update.yml
```diff
@@ -0,0 +1,13 @@
+name: "Update Angular Action"
```
floogulinc, line 1:
```suggestion
name: "Update Angular"
```

floogulinc (changes requested):

.github/workflows/ng-update.yml
NicMcPhee: Good catch – thanks!

helloworld12321 (approved):
Cool! 🙂

floogulinc (approved):

PR #1168, opened 2022-09-01
Fix ng update action

floogulinc:


NicMcPhee (approved):
Thanks for catching that! This is what I get for copy/pasting straight from their docs without thinking.

PR #1170, opened 2022-09-01
Set the project path for the Angular update action

floogulinc:


NicMcPhee (approved):
We'll get there eventually...

PR #1279, opened 2023-02-01
Update recommended extensions and some documentation

### LINKED ISSUE #1033: Add Thunder Client to recommended extensions
This would actually be very useful from lab 2 on. Thunder Client is a lightweight REST client for VS Code similar to things like Insomnia and Postman.

https://www.thunderclient.com/

https://marketplace.visualstudio.com/items?itemName=rangav.vscode-thunder-client

kklamberty:
This branch was created initially to update the recommended extensions to include Thunder Client as suggested in #1033. I believe that work on this branch has added the extension and updated the documentation associated with the Iteration Template appropriately. I also did remove the badges for BCH and LGTM here since I was touching the documentation and it didn't seem inappropriate to take care of that.

Closes #1033


DEVELOPMENT.md
```diff
@@ -161,6 +161,8 @@ From the `server` directory, `./gradlew test` runs the server tests once.
     to the regular report generated by the `test` task.
   - It will fail if any of the the test coverage metrics is under 80%.
 
+In addition to these automated server tests, you might want to manually explore the requests and different parameters at the API level. To see what is happening and explore your API, you can use [Thunder Client](https://www.thunderclient.com/). There are more instructions about how to do this in [Lab 2](https://github.com/UMM-CSci-3601/3601-lab2_client-server/blob/main/LABTASKS.md#use-thunder-client-to-explore-api-output).
```
NicMcPhee, line 164: I wouldn't consider this blocking (maybe a new issue?), but I think it's risky to link back to previous labs. What I'd recommend instead is taking all @jpwalbran's work (screenshots, etc.) and move them into their own file. Then I'd probably put that file _here_ and have the labs link _forward_ to that. Or we just copy the file across the various labs.

Related (and, again, this could be a new issue), I'd extend them to say something about `PUT` and `POST`, because that's where Thunder Client is most useful. If all you're doing is `GET` (which is all you _can_ do in Labs 2 and 3), then you could just hack URLs. When you want to do more "interesting" things like `POST` then it's really useful to have something like Thunder Client that can generate HTTP request bodies for us (which you can't do by hacking URLs).

NicMcPhee (approved):
See my previous review.

DEVELOPMENT.md
kklamberty: I think you're right that it would be great to make a more complete reference/guide with hints about a variety of requests. I noted that there was a link back to Lab 0 in here somewhere and thought it might be a good idea, but thinking back... that reference is not always accurate either. Where should a file like that exist? Should we refer to that file from multiple places? I'm not sure what repository it should live in, but I would suggest that the Iteration Template may be the best place.

NicMcPhee: I do think that if we're going to have shared files, they should be here, and labs should link forward to them. I think that maximizes the likelihood of consistency.

Ultimately it would be nice to have a monorepo setup so we would have just one copy of something like that, and it would deploy out to all the different repos when we make changes, but that's complicated and not something that's happening this semester.

NicMcPhee (approved):
Thanks for bringing @jpwalbran's docs on Thunder Client into this repo.

I do think we still want at least an issue about extending that documentation to include things like `POST` requests.

kklamberty:
I've added issue #1282 in response to discussion here. Thanks!

Issue #1246, opened 2022-11-17
Replace @angular/flex-layout with css flexbox


floogulinc:
The [@angular/flex-layout](https://github.com/angular/flex-layout) package is being deprecated and will receive its last update for Angular 15. 

See [this article](https://medium.com/@caerus.karu/farewell-flex-layout-aaa567023769) for more.

kklamberty:
Here are some guides for doing this:

- Somewhat official guide: https://github.com/angular/flex-layout/issues/1426
- Includes a link to the somewhat official guide: https://github.com/angular/flex-layout/issues
- One person's approach [here](https://github.com/angular/flex-layout/issues/1426#issuecomment-1353246488) shows some changes and the accompanying css.

PR #1280, opened 2023-02-01
Update to Angular 15 and Javalin v5

jpwalbran:
The main purpose of this PR is to go through the update process to Angular 15, and then adjust anything that comes up as necessary. In the process of doing that, there were some other big things that came up and wound up in this PR.

This PR now includes:

- Updating Angular from 14 to 15
- Removing the Angular flex-layout
- Updating Angular Material to the newer MDC specifications (and updating Cypress e2e tests as needed to address these changes)
- Updating from Javalin v4 to v5 (and updating the tests for UserController)

The change to Javalin v5 was substantial, and required quite an overhaul of the tests. The UserControllerSpec in the iteration template matches more with what the students did in Lab 2 and includes use of Mockito as a replacement for the ContextUtil (which went away as a result of Context becoming an interface in v5 of Javalin). The challenges of matching actual usage of Javalin in the Mockito tests were significant. We probably do want to consider whether or not we like this type of test or if we would like to move to more functional testing like @NicMcPhee did with Unirest in this other branch of another repo: https://github.com/UMM-CSci-3601/3601-lab2_client-server/pull/224

This PR leaves the client looking a bit scraggly, but there is a PR in motion for fixing that somewhat: #1287


kklamberty:
Ideally, we would also make the Angular stuff look a bit nicer before we start splitting out the elements for the labs. This is especially true for lab 3. We've removed flex-layout, but we haven't really dealt completely with the fallout of that choice. For example, the user list view is not looking particularly pretty yet. 

One reason to go ahead and merge in this branch is that is is focused on the updating of versions. We could have other branches that focus on separate details like making some changes that either don't rely on flex-box or could replace flex-box. There are some specifics about that in issue #1246 that are really not addressed here.

client/cypress/support/app.po.ts
```diff
@@ -15,7 +15,7 @@ export class AppPage {
     return cy.get('.sidenav');
   }
 
-  getNavLink(navOption: 'Home' | 'Users') {
-    return cy.contains('[routerlink] > .mat-list-item-content', `${navOption}`);
+  getNavLink(navOption: ' Home ' | ' Users ') {
```
NicMcPhee, line 18: Why do you put spaces around these? That seems odd, and not something I would expect people (students) to do in the future? Is there a reason for it?

client/package.json
```diff
@@ -13,32 +13,31 @@
...
     "rxjs": "~7.5.5",
     "tslib": "^2.4.0",
     "zone.js": "~0.11.6"
   },
   "devDependencies": {
-    "@angular-devkit/architect": "^0.1402.4",
-    "@angular-devkit/build-angular": "^14.2.4",
+    "@angular-devkit/architect": "^0.1501.3",
+    "@angular-devkit/build-angular": "^15.1.3",
     "@angular-eslint/builder": "~14.1.2",
```
NicMcPhee, line 33: We need to update the `@angular-eslint` stuff to v15 as well, don't we?

client/package.json
```diff
@@ -13,32 +13,31 @@
   },
   "private": true,
   "dependencies": {
-    "@angular/animations": "^14.2.4",
-    "@angular/cdk": "^14.2.3",
-    "@angular/common": "^14.2.4",
-    "@angular/compiler": "^14.2.4",
-    "@angular/core": "^14.2.4",
-    "@angular/flex-layout": "^14.0.0-beta.40",
```
NicMcPhee, line 21: Nice job getting rid of `flex-layout`.

client/src/app/home/home.component.html
```diff
@@ -1,6 +1,6 @@
 <div fxLayout="row">
   <div fxFlex fxFlex.gt-sm="80" fxFlexOffset.gt-sm="10">
-    <mat-card class="home-card">
+    <mat-card appearance="outlined" class="home-card">
```
NicMcPhee, line 3: Was this (and all the other additions of `appearance="outlined"`) needed as part of the update, or was it seen as an improvement to the interface? (Without running the previous version next to this one, it's hard to see what [if anything] this does.)

client/src/app/users/user-list.component.scss
```diff
@@ -12,7 +12,7 @@
 }
 
 .mat-radio-button {
-  margin: 0 12px;
+  margin: 0 1rem;
```
NicMcPhee, line 15: Nice change. Do we use `px` anywhere else?

client/src/test.ts
```diff
@@ -7,13 +7,6 @@ import {
   platformBrowserDynamicTesting
 } from '@angular/platform-browser-dynamic/testing';
 
-declare const require: {
```
NicMcPhee, line 10: Why don't we need this anymore? Is the default (somewhere) now to pick up all `.spec` files?

server/src/main/java/umm3601/user/User.java
```diff
@@ -34,4 +34,19 @@ public class User {
...
+  @Override
+  public boolean equals(Object obj) {
+    if (!(obj instanceof User)) {
+      return false;
+    }
+    User other = (User) obj;
+    return _id.equals(other._id);
+  }
+
+  @Override
```
NicMcPhee, line 47: Good catch on the need to override the `hashCode` – sorry I missed that.

server/src/main/java/umm3601/user/UserController.java
```diff
@@ -93,23 +98,33 @@ public void getUsers(Context ctx) {
...
   }
 
   private Bson constructFilter(Context ctx) {
     List<Bson> filters = new ArrayList<>(); // start with a blank document
 
     if (ctx.queryParamMap().containsKey(AGE_KEY)) {
-        int targetAge = ctx.queryParamAsClass(AGE_KEY, Integer.class).get();
-        filters.add(eq(AGE_KEY, targetAge));
+      int targetAge = ctx.queryParamAsClass(AGE_KEY, Integer.class)
+        .check(it -> it > 0, "User's age must be greater than zero")
```
NicMcPhee, line 115: Nice job adding the validation to these requests.

NicMcPhee (changes requested):
I really don't think you should merge this (as it claims to be about updating to Angular v15) until you also update the Angular `eslint` bits to v15.

My other comments are more cosmetic, or just straight up questions.

I agree 100% with @kklamberty's comment that we should _not_ let things like tidying up the appearance block this PR. Things like that are important, but should be in different PR(s) so we can get this merged in.

NicMcPhee:
You should also update the PR description so that it more accurately reflects what all is actually happening in this PR, as it's grown well beyond just upgrading Angular to v15.

client/src/app/users/user-list.component.scss
kklamberty: I think we use px *everywhere* else. I tried this kind of on a whim...

client/package.json
kklamberty: I'm not convinced this is actually ideal. The appearance of the cards and the lists is quite messed up by it. I was *just* about to make a commit that goes back to flex-layout so that we at least have cards that look decent. The list is still very broken either way. I'm not convinced that I have enough time before Lab 3 on *Tuesday* to do anything very good here. @NicMcPhee - Thoughts?

server/src/main/java/umm3601/user/User.java
kklamberty: No worries!

client/package.json
NicMcPhee: I upgraded this in [50ad4ad](https://github.com/UMM-CSci-3601/3601-iteration-template/pull/1280/commits/50ad4adae0c981cfe7fa9a60d0f5ccae28049c5f), so that's good now.

client/cypress/support/app.po.ts
kklamberty: I wonder if this happened with some of the automated tooling for moving to Angular 15? Or, possibly @jpwalbran made a choice here?

kklamberty (approved):

kklamberty (dismissed):

client/src/app/home/home.component.html
kklamberty: It did look a little bit like you could not discern the location of the edges of the cards when we removed (and had not replaced) flex-layout, so I suspect this explains the reasoning behind this change.

client/src/app/users/user-list.component.scss
floogulinc: The Angular Material stuff generally uses px values for styles are least as of the last time I checked.

.github/workflows/e2e.yaml
```diff
@@ -35,10 +35,10 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-gradle-
 
-    - name: Use Node.js 16
+    - name: Use Node.js 18
```
floogulinc, line 38: If you're switching to Node 18 here then you should update it in the other workflows and the Dockerfile.

floogulinc:
In the future when using `ng update` please use the `--create-commits` flag so it creates separate commits for each migration. This makes it much easier to see what was changes and why since each one has a comment explaining the change.

NicMcPhee:
I did that, for example, on 50ad4adae0c981cfe7fa9a60d0f5ccae28049c5f. To be fair, those commit messages aren't amazing and don't read super well on GitHub, but they're consistent and do list all the updates, which is nice.

kklamberty:
> In the future when using `ng update` please use the `--create-commits` flag so it creates separate commits for each migration. This makes it much easier to see what was changes and why since each one has a comment explaining the change.

Good call! I guess we just followed the instructions on the page about the migration and didn't also add that, but I will try to remember that for the future. Thanks for for the note!

.github/workflows/e2e.yaml
kklamberty: Good catch! I definitely would not have remembered the Dockerfile.

kklamberty:
Hey, All! I think I've addressed the changes requested. @NicMcPhee - can you please re-review? We need an approval from you to do this merge.

NicMcPhee (approved):
Thanks for all the good work - this was a bit of a bear.

PR #1287, opened 2023-02-06
Try non flex-layout

kklamberty:
This branch is a place to try out some things to ease ourselves away from the `flex-layout` that is no longer going to be updated. (See #1246)

The work done in this PR for moving away from `@angular/flex-layout` was accomplished by following an idea here: https://github.com/angular/flex-layout/issues/1426#issuecomment-1304100298

Then, I made some additional changes so that the list view looks okay.

I'm not sure this is the best approach to fixing this problem. Since things will be supported for about a year, we *may* want to see about undoing the much smaller change made in a previous commit that removed the flex-box dependency to see what would break. 0016464a7347023e27490844bda54522455a6cf9

There is also something in 19c287e718b3a20360a360fc4349e6843afdea86 that we need if we want to go back to the flex-layout for now.


floogulinc:
I assume you mean moving away from @angular/flex-layout not flex-box since you should still be having flex-box to replace it

README.md
```diff
@@ -37,7 +37,7 @@ From the `server` directory:
 - `./gradlew test` to test the server
 - `./gradlew checkstyleMain` to run Checkstyle on the server Java code in the `src/main` folder
 - `./gradlew checkstyleTest` to run Checkstyle on the server Java code in the `src/test` folder
-- `./gradlew check` will run both the tests and Checkstyle in one command
+- `./gradlew check` will run the tests, run the Checkstyle, and generate coverage reports in one command
```
NicMcPhee, line 40:
```suggestion
- `./gradlew check` will run the tests, run the Checkstyle checks, and generate coverage reports in one command
```

NicMcPhee: Note that these changes have nothing to do with `@angular/flex-layout` and probably should have been in another PR. 😜

client/cypress/e2e/user-list.cy.ts
```diff
@@ -49,23 +49,19 @@ describe('User list', () => {
...
 
-    // Go through each of the cards that are being shown and get the companies
-    page.getUserCards().find('.user-card-company')
-      // We should see these companies
-      .should('contain.text', 'MOMENTIA')
-      .should('contain.text', 'KINETICUT')
-      // We shouldn't see these companies
-      .should('not.contain.text', 'DATAGENE')
-      .should('not.contain.text', 'OHMNET');
+    // Each user card's company name should include the text we are filtering by
```
NicMcPhee, line 54: Nice improvement.

NicMcPhee (dismissed):
This is a definite improvement – thanks! 

I had one small editing change that I think should probably be done, but you could also do that somewhere else (like a PR focusing on updating the write-up).

I wondered if the CSS should be in a more "specific" place (like connected to the user list), but it's all pretty general and might be useful in other places, so I guess it's OK where it is?

NicMcPhee (approved):
Thanks!

PR #1289, opened 2023-02-07
Pull forward some lab elements (maybe)

kklamberty:
This PR consists of mostly minor changes that reflect differences I noticed between Lab2, Lab3, and the Iteration Template while I was in the process of making Lab 3 for Spring 2023.


client/cypress/support/user-list.po.ts
```diff
@@ -42,6 +65,7 @@ export class UserListPage {
     return cy.get('[data-test=userRoleSelect]').click()
       // Select and click the desired value from the resulting menu
       .get(`mat-option[value="${value}"]`).click();
+      // NOTE: THIS CHAINING MIGHT BE FRAGILE (due to a 'click' followed by a 'get')
```
jpwalbran, line 68: It seems to be fine for me, even when I run the test a few times with a throttled network.

client/src/styles.scss
```diff
@@ -113,6 +113,7 @@ body {
   text-transform: uppercase;
 }
 
+/////////// Remove flex-layout and replace with CSS ////////////////////
```
jpwalbran, line 116: I'm not sure how long we need this to live on the project. It definitely makes sense now, but it's probably going to only be confusing in a year or two. (Or to students who aren't paying attention to old iteration templates, and this might just be confusing for them.)

jpwalbran (changes requested):
It mostly looks good, there's just one thing that I think should at least be decided on before merging in. Probably a pretty simple change (or not a change at all).

client/cypress/support/user-list.po.ts
jpwalbran: If it does turn out to be fragile, it's probably sufficient to add some form of visibility check to the test before the second get, just to make sure that it tries a few times before giving up on the second get().

kklamberty: Most of the time, I think that the e2e tests fail on GitHub even if they don't fail elsewhere. When you were in Software D last year, was that your experience? Or, were you often able to see the tests fail in the lab too? My experience this semester with e2e tests has only been from home since I have not been to campus.

It's also the case that the problem of detaching from the DOM has been mentioned as being much less in this newest version of Cypress that we upgraded to just a little bit before Lab 3 this semester, so maybe it's less fragile.

jpwalbran: That very well may be the case. I remember the DOM detachment happening a lot on both, but I suppose it was more common on the remote. My hope was that artificially throttling my network (from the Chrome debug tools), would be sufficient to maybe create a race condition that could possibly cause a DOM detachment error or similar. My testing was far from thorough though, so keeping this there. Additionally, adding visibility checks was what I finally found sufficient to solve the race-condition problems.

kklamberty: Cypress has some nice documentation about how to avoid this kind of fragile chaining, but I thought it might be worthwhile to keep in a warning. I guess a potentially better approach might be to add a mention about what to do if the tests seem fragile? Should we create a new issue that specifically addresses how we want to provide that kind of support (and words of warning/wisdom) for students about Cypress tests? I think this blog post is a great choice for where to point people who are experiencing fragile DOM stuff: https://www.cypress.io/blog/2020/07/22/do-not-get-too-detached/ What do you think would best support the students?

client/src/styles.scss
kklamberty: Good point. I'm not sure either. I wish there was a way to put a giant post-it note on it so we would notice, but it would not show up for students. Since we are still a bit "in process" for replacing flex-layout, I think we should leave it *for now* and plan to remove it. Maybe there should be an issue devoted to removing these types of hints that might be useful only in the short term... and a way to collect what things those are?

client/cypress/support/user-list.po.ts
kklamberty: Ok - I removed the comment and attempted to fix the fragile chaining so that it would not be fragile anymore.

client/src/styles.scss
kklamberty: I think we should leave this in for now.

kklamberty (commented):
Based on discussions with @floogulinc, this commit a363fd25c2175b958202390dea1003fc20dbc41c may be problematic. We should probably be careful here. There is a conversation on that commit that may help.

client/src/styles.scss
jpwalbran: Got it.

jpwalbran (approved):
LGTM!

Issue #793, opened 2022-03-14
Fix case insensitive queries so they are actually case insensitive


NicMcPhee:
Our case insensitive queries in the example `UserController` code aren't actually case insensitive, which came as a non-trivial surprise this semester. Despite what lots of things say on this Internet, adding `"i"` as a `Pattern` property like this:

```java
      filters.add(regex(COMPANY_KEY,  Pattern.quote(ctx.queryParam(COMPANY_KEY)), "i"));
```

doesn't actually work.

@miiiiip did some digging and shared a solution on Slack, which is to add a `?` in front of the `i`:

```java
      filters.add(regex(COMPANY_KEY,  Pattern.quote(ctx.queryParam(COMPANY_KEY)), "?i"));
```

I honestly don't get why we need `"?i"` instead of just `"i"`, but I'm sure it's in some documentation somewhere.

I also tried using the `Pattern.CASE_INSENSITIVE`, but never got that to work. I was pushing pretty fast, though, so it might be worth revisiting that. [This post on "Case-insensitive string matching in Java"](https://www.baeldung.com/java-case-insensitive-string-matching) might be a useful resource.

This also should be applied to [Lab 4](https://github.com/UMM-CSci-3601/3601-lab4-mongo/issues/43).

NicMcPhee:
So I think the problem is that we were basically just adding `i` to the search string, and we needed what they call an "embedded flag expression `(?i)`" to indicate that we were changing "modes" to use case-insensitive searching. Apparently you can do fancy stuff with that to change case-sensitivity locally in subsets of a pattern.

TBH, I think the better solution for us is to use the `CASE_INSENSITIVE` when we compile the `Pattern` rather than futzing about with the rather obscure `(?i)` notation. [The Oracle Java docs cover this in a fairly reasonable way](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/regex/Pattern.html#CASE_INSENSITIVE).

kklamberty:
There are a couple of things I'm just now noticing:
We give examples of filtering on the client (`company` and `name`) and filtering on the server (`age` and `role`). We label those on the client side to help clarify where we are going to do that work. However, we *actually* have something that *looks* like it's going to filter the `company` on the server, and it seems to be the case that the way that flag "i" and the pattern were set up was not working. 

[...]

PR #1290, opened 2023-02-11
Update query and approach to ensure case insensitivity

kklamberty:
Addresses #793 in which a bug is reported in the filtering. I am not able to recreate this bug, and the filtering seems to handle queries in a case-insensitive way. But, this approach at least takes some of the mystery out of what the "i" was for, which could be helpful for anyone less familiar with regex concepts.

This PR focuses on making sure that the query to the database will check the pattern in a case-insensitive way. The old approach seems to work, but this approach makes the case insensitive option very clear and uses Pattern in a conventional way.


kklamberty:
There are a couple of things I'm just now noticing:
We give examples of filtering on the client (`company` and `name`) and filtering on the server (`age` and `role`). We label those on the client side to help clarify where we are going to do that work. However, we *actually* have something that *looks* like it's going to filter the `company` on the server, and it seems to be the case that the way that flag "i" and the pattern were set up was not working. 

I don't think that this would come up as not working unless someone removed the filtering from the client side. I *suspect* that this shows up more if they attempt to use our server side "filtering" by company as a model for other work like filtering todos, that is where they run into a problem. This might be why I was not able to recreate this bug.

Do we actually want to leave this filtering by company on the server at all? It seems like we should remove all things resembling an attempt to filter company on the server since we claim to be doing so on the client side of things. On the other hand, having a model that includes a regex (used correctly) might be handy.

Thoughts?

kklamberty:
It's not possible to enter anything other than the legal options for role exactly as they are typed because of the use of the ROLE_REGEX and checking of the validity of the http parameters against that very strict set of roles. So, there isn't even anything interesting that Thunder Client can tell us since the regex in place for validating the parameter is already so tied down. That didn't used to be the case, though. So, perhaps this is not a concern anymore?

kklamberty:
Ok. I can verify that if we make it so that only name is filtered by the client, the regex with case insensitive matching as entered in this PR works to do company searching without paying attention to case. To really know that was the case, I needed to:

1. Remove filtering by company from the client side (two places in `user.service.ts` and in the `updateFilters` method in `user-list.component.ts`)
2. Add `company: this.userCompany` in `this.userService.getUsers`
3. Update the regex stuff as shown in this PR

HOWEVER, it still seems to do case-insensitive queries on Thunder Client with the original option, the ?i option AND the one I propose here. So, I really don't know what we want to do. I do not know how to recreate this bug... only that the three options for case insensitive searching we have suggested all work in Thunder Client.

kklamberty:
Also, [this page's example](https://www.baeldung.com/java-case-insensitive-string-matching#pattern) seems to suggest that we want both `Pattern.quote` and `Pattern.compile` someplace in our solution...

NicMcPhee:
This does appear to fix the problem If I enter `GET` requests in the browser [like I did here](https://github.com/UMM-CSci-3601/3601-lab4-mongo/pull/54#issuecomment-1427202220) then this PR works.

As far as where the filtering should happen, I think it's entirely reasonable to have filtering by company here, even if "officially" we're doing company filtering on the client side for several reasons. It is, as you said, useful to have the REGEX examples here. It's also useful to have a variety of features (including filtering by company) even if you don't ever call them from the web UI. Maybe there's some other tool using the server that does use that, for example, like a server-to-server system. So just because our client doesn't every call/use this feature doesn't mean it has no potential value.

We can "officially" test this by mocking HTTP requests on the server side and we can test it "by hand" through `GET` requests in the browser or through Thunder Client. Does that make sense?

NicMcPhee (dismissed):

NicMcPhee:
So I think this doesn't need to happen, as the probably appears to have magically gone away here as it did on [Lab 4](https://github.com/UMM-CSci-3601/3601-lab4-mongo/issues/43).

That said, I do suspect it might be a Good Idea to use the `.compile()` command, although TBH I haven't looked into that.

So I think we should either:

- Just close this and move on, or
- Rename this to be something about improving our use regex `Pattern`, and also move those improvements back to Lab 4.

Thoughts @kklamberty?

kklamberty:
I think that we should update the use of the regex `Pattern` both here and in lab 4. But, perhaps that is a different issue and not really part of fixing the bug that seems to have wandered off somehow. I think we might want to use both compile and quote as I have seen that done elsewhere, so maybe not exactly as I've written here.

NicMcPhee (approved):
I think this looks great. I don't know that it actually solves the problem advertised in the title (since I'm still not really sure where that actually got solved). If you wanted to you could change the name and description of this PR if you think that could improve the likelihood of someone finding this PR? Not sure what you'd change it _to_, though.

That said, I think the code looks wonderful, and we should totally add these changes here and in Lab 4.

kklamberty:
> I think this looks great. I don't know that it actually solves the problem advertised in the title (since I'm still not really sure where that actually got solved). If you wanted to you could change the name and description of this PR if you think that could improve the likelihood of someone finding this PR? Not sure what you'd change it _to_, though.
> 
> That said, I think the code looks wonderful, and we should totally add these changes here and in Lab 4.

I think [this Stack Overflow answer](https://stackoverflow.com/a/15409401) helps me feel a little better about how to ensure that the search pattern is set up nicely now. I *did* change the name from the original name, but I agree that it's still imperfect.

Issue #349, opened 2021-03-03
Clean up the subscriptions in the `UserListComponent`


NicMcPhee:
Our use of `subscribe` and `unsubscribe` in `UserListComponent` is ugly. We'd like to move more to pipes and maps and perhaps proper debouncing. This comes from good ideas from @floogulinc and @helloworld12321.

This should apply backwards, probably as far as Lab 2.

floogulinc:
I have started work on this in https://github.com/UMM-CSci-3601/3601-iteration-template/commit/e132ebb28e2b7f35f30bf02752fd3e195ba5d4d1

floogulinc:
I wrote this as a comment on a commit but I'll put this here too in case it helps. Here is an example of refactoring to better use rxjs techniques in the user profile component:
```ts
export class UserProfileComponent implements OnInit {
  id$: Observable<string> = this.route.paramMap.pipe(
    map((pmap) => pmap.get('id'))
  );
  user$: Observable<User> = this.id$.pipe(
  	switchMap((id) => this.userService.getUserById(id)),
    shareReplay(1) // allows multiple subscriptions without making multiple requests
  );
  constructor(private route: ActivatedRoute, private userService: UserService) { }
}
```
Then switch out plain uses of `user` and `id` with `user | async` and `id | async` in the template. 
I wrote this in the GitHub comment editor so please excuse any slight mistakes.

floogulinc:
I'm going to reopen this as the changes still aren't really what this issue is asking for. I think the goal would be to remove the use of manual subscribe all together in favor of a more reactive approach using the async pipe in the template.

kklamberty:
This is quite unlikely to happen for Spring 2023 in the absence of a PR that actually handles the work that is requested.

kklamberty:
Relates (potentially) to work done for Lab 4 [in this PR](https://github.com/UMM-CSci-3601/3601-lab4-mongo/pull/68)

kklamberty:
In case it's ever helpful to have an example of [adding debounce behavior](e1debdb6d9afb9916e5f5c1ff78d8d0aa78f7bc9) and [updating tests to pass with that new behavior](afedc98df8c1a9876a327c6a1df00fe2d6161c64), here are commits on [a branch/PR](https://github.com/kklamberty/three-tined-spork/pull/10) where I did that.

PR #1291, opened 2023-02-11
Change user-list component's use of subscription

kklamberty:
The way we were using subscribe was deprecated. Its signature involved two inputs to the subscribe, one of which was the error, and that signature is not recommended or supported anymore. Instead, this uses a structure that tells what to do when the `next` change is emitted from the observable, what to do for `error`, and what to do when changes to the observable are `complete`. 

I also added some comments to help describe what's happening here since observables and asynchronous behaviors are new to most students in the course, and it's tough to understand or know where to look for more description about it.

### DIFF
--- a/client/src/app/app.module.ts
+++ b/client/src/app/app.module.ts
@@ -6,7 +6,7 @@ import { AppComponent } from './app.component';
 import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
 import { FormsModule, ReactiveFormsModule } from '@angular/forms';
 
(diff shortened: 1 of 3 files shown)


kklamberty:
This partially addresses #349 because it changes our use of subscribe, but #349 also says we might use pipe, map, and debouncing... and my attempts to do that along with the subscribe were not working properly. I was able to use pipe and debounce, but it wasn't actually changing the behavior so I removed it thinking that I must not have been using it correctly.

PR #1293, opened 2023-02-12
Alphabetize imports and lists of modules

kklamberty:
It's tough to spot differences between long lists of imports or notice if anything is missing. I'm alphabetizing these to make it a little easier.

### DIFF
--- a/client/src/app/app.module.ts
+++ b/client/src/app/app.module.ts
@@ -53,22 +53,22 @@ const MATERIAL_MODULES: any[] = [
 
 @NgModule({
   declarations: [
-    AddUserComponent,
     AppComponent,
     HomeComponent,
-    UserCardComponent,
     UserListComponent,
+    UserCardComponent,
     UserProfileComponent,
+    AddUserComponent,
   ],
   imports: [
+    BrowserModule,
     AppRoutingModule,
     BrowserAnimationsModule,
-    BrowserModule,
     FormsModule,
-    HttpClientModule,
-    LayoutModule,
-    MATERIAL_MODULES,
     ReactiveFormsModule,
+    HttpClientModule,
+    MATERIAL_MODULES,
+    LayoutModule,
   ],
   providers: [
     UserService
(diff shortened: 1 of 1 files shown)


client/src/app/app.module.ts
```diff
@@ -1,74 +1,74 @@
...
+    AddUserComponent,
     AppComponent,
     HomeComponent,
-    UserListComponent,
     UserCardComponent,
+    UserListComponent,
     UserProfileComponent,
-    AddUserComponent,
   ],
   imports: [
```
floogulinc, line 63: The order of these matters. I would be careful arbitrarily reordering them like this. If two modules have providers for the same token, the later one wins. Also routing order is affected by the order of routing modules here.

NicMcPhee:
Maybe it would be more useful to group these into combinations of related modules and then all you import is those groups. That would probably improve readability and reliability, and if we're careful we should be able to avoid the sequencing problems that @floogulinc brought up.

kklamberty:
Which are the parts where it matters? And, how can we uncover where it matters? I mainly ask because it feels like when you add a new one, where it gets added has not been obviously meaningful in my experience. Can the imports be alphabetical at the top, but we have to have some other order for the imports in the @NgModule? Does the order in the declarations part matter? Where can I read more? This looked fruitful, but it really doesn't help me figure out the order: https://angular.io/guide/ngmodule-faq. This also looked promising, but again... no real clues about how to *know* what order to use: https://angular.io/guide/bootstrapping#launching-your-app-with-a-root-module. Is it enough to know that the tests pass?

floogulinc:
The order doesn't matter at all for the typescript imports at the top of the file and not for the `declarations` section of the module. You should also be fine writing the Angular Material imports in `MATERIAL_MODULES` in any order since they shouldn't have any conflicts. I think it is mainly within the `imports` section of the NgModule declaration that it matters. It may be the case that none of the things we import have any conflicting providers but it's possible so I just wanted to warn about that. 

Also in general I think it's good to keep the BrowserModule as the first thing in that list, it looks like you don't _have_ to but it acts as the base of all Angular apps so it feels better to have it first imo.

NicMcPhee (approved):

Issue #1282, opened 2023-02-01
Update documentation on Thunder Client to describe POST requests


kklamberty:
We have a nice start on documentation for Thunder Client [here](https://github.com/UMM-CSci-3601/3601-iteration-template/blob/main/THUNDER_CLIENT.md), but it would be good to add some further examples including more than GET requests. Having this type of documentation in the Iteration Template repo seems like a good place, and if there are other tools to document at all, other files could be made here. We probably want to refer forward to the documentation here from the other labs (like Lab 2).

NicMcPhee:
I think it's particularly important to have an example of a `POST` or `PUT` request that puts info in the body of the request. You can "fake" all `GET` requests by constructing URLs, but things that need a body can't be done that way, and that's where something like Thunder Client really shines.

PR #1294, opened 2023-02-13
Update Thunder client documentation for POST/PUT

jpwalbran:
Added instructions for making POST requests with ThunderClient, using the example addUser request.

### DIFF
--- a/THUNDER_CLIENT.md
+++ b/THUNDER_CLIENT.md
@@ -6,7 +6,7 @@ really helpful when you're trying to debug what your server gives you.
 
 ## Use Thunder Client to explore API output
 
-To use Thunder Client (once it's installed), open it from the sidebar.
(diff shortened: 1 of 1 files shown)


jpwalbran:
Closes #1282

NicMcPhee (dismissed):
Thanks for this very helpful addition!

This has gotten long enough that I'd add a table of contents to the top. These can be generated and managed by the "Markdown All In One" VS Code extension. I don't consider that a blocking change, but it would be nice to do.

Also, if it wouldn't stress KK too much, this would be nice to also add to the Lab 4 docs.

kklamberty:
Thanks, @jpwalbran , for creating this documentation! I've made the change @NicMcPhee suggested to add a table of contents, and I've also made a few other minor changes.

kklamberty (approved):

PR #1297, opened 2023-02-23
Fix Deep Source suggestions

kklamberty:
This PR focuses on fixing suggested problems noticed by Deep Source. It's important to note that approving this PR *does not* also include a new check. I used the check along the way to help me notice things to fix, but I did not leave the configuration fine here. I put that in a separate PR. This PR should be considered separate from the adoption of Deep Source.


floogulinc:
Can you explain the reasoning behind moving all the selectors and other strings in the cypress files to static members of the class? I don't think they're reused very much if at all and moving them away from where it's used seems to just add clutter.

floogulinc:
Also in general do you plan on having students use this tool? Where do you even see the recommendations? If you have to sign into their panel then the 3 user limit on the free plan seems like it will be an issue.

.deepsource.toml
```diff
@@ -0,0 +1,16 @@
...
+  runtime_version = "11"
+
+[[analyzers]]
+name = "shell"
+
+[[analyzers]]
+name = "javascript"
+
+  [analyzers.meta]
+  plugins = ["angular"]
```
floogulinc, line 16: I've made some changes to the config based on [their docs](https://deepsource.io/docs/). 

```suggestion
version = 1

test_patterns = [
  "*.spec.ts",
  "client/cypress/**",
  "server/src/test/**"
]

[[analyzers]]
name = "java"

  [analyzers.meta]
  runtime_version = "11"

[[analyzers]]
name = "shell"

[[analyzers]]
name = "javascript"

  [analyzers.meta]
  plugins = ["angular"]
  dependency_file_paths = [
    "client/",
  ]

[[analyzers]]
name = "docker"
enabled = true

  [analyzers.meta]
  dockerfile_paths = [
    "client/Dockerfile",
    "server/Dockerfile"
  ]
```

kklamberty:
> Can you explain the reasoning behind moving all the selectors and other strings in the cypress files to static members of the class? I don't think they're reused very much if at all and moving them away from where it's used seems to just add clutter.

I moved them so that the methods would use `this`. I don't mind not utilizing that code quality check, but it did cause me to think about what the purpose was of having the methods be instance methods. I never really thought about that before. They are not static in the class sense like Java, but they are unchanging.

kklamberty:
> Also in general do you plan on having students use this tool? Where do you even see the recommendations? If you have to sign into their panel then the 3 user limit on the free plan seems like it will be an issue.

It sounds like the number of users won't be a problem for public repositories for people with the GitHub Student Developer pack.

kklamberty:
The update suggested by @floogulinc in the previous commit brings up a problem in Deep Source about how we use CMD in docker. Here is some information on it from https://docs.docker.com/engine/reference/builder/#cmd: 
>If you want to run your <command> without a shell then you must express the command as a JSON array and give the full path to the executable. This array form is the preferred format of CMD. Any additional parameters must be individually expressed as strings in the array

This comment is in relation to the line of code in the server dockerfile that says `CMD /wait && server/bin/server`

kklamberty:
It seems likely that the ideas suggested by Deep Source and the decision to *use* Deep Source might be better handled on separate branches and with separate PRs.

jpwalbran (approved):
LGTM!

PR #1299, opened 2023-02-23
Consider adding Deep Source as a code quality check

kklamberty:
This PR is a place to consider adding Deep Source as a code quality check

### DIFF
--- a/.deepsource.toml
+++ b/.deepsource.toml
@@ -1,0 +1,35 @@
+version = 1
+
+test_patterns = [
+  "*.spec.ts",
+  "client/cypress/**",
+  "server/src/test/**"
+]
+
+[[analyzers]]
+name = "java"
+
+  [analyzers.meta]
+  runtime_version = "11"
+
+[[analyzers]]
+name = "shell"
+
+[[analyzers]]
+name = "javascript"
+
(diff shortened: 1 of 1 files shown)


kklamberty:
Originally posted by @floogulinc in [this commit](https://github.com/UMM-CSci-3601/3601-iteration-template/pull/1297/files/da37762d5fce67585f6793e04b4cbfc254c25d2d#r1115319596)
> I've made some changes to the config based on [their docs](https://deepsource.io/docs/).

In a separate PR #1297

kklamberty:
I found many of the suggestions Deep Source gave me helpful and informative. A few things might have caused me to do things I didn't really want to do or need to do, and I have slight concerns that students might respond to those things by applying them without knowing it's okay to push back a little... or ignore things they shouldn't... but I think in all it's more helpful than not, and including it might be useful.

NicMcPhee:
I'm cool with bringing this in; I thinks it's a reasonable replacement for a lot of what LGTM was doing for us.

I still think there's some interesting stuff that BetterCodeHub checked for us that we don't have a replacement for, especially copy/paste detection, but that's for another issue not this one.

NicMcPhee (approved):

PR #1300, opened 2023-02-23
Increase coverage of `User.java`

kklamberty:
I'm not sure this is really great testing, but it does increase the coverage.

### DIFF
--- a/server/src/main/java/umm3601/user/UserController.java
+++ b/server/src/main/java/umm3601/user/UserController.java
@@ -36,6 +36,7 @@ public class UserController {
   static final String AGE_KEY = "age";
   static final String COMPANY_KEY = "company";
   static final String ROLE_KEY = "role";
+  static final String SORT_ORDER_KEY = "sortorder";
 
   private static final int REASONABLE_AGE_LIMIT = 150;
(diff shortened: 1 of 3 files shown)


jpwalbran (approved):
Looks good.

NicMcPhee (changes requested):
So this does improve the coverage, but I think it does it in the wrong place. What you're trying to do here is test the the `equals()` and `hashCode()` methods in `User`. That really should be in a new `UserSpec.java`, not being done indirectly here in the controller tests.

kklamberty:
Oh, that totally makes sense. What I really wanted to do was use the equals method, but it was just not happening. You're right. What I actually ended up doing was something that was not a test of the controller, but rather a test of the User class. Thanks!

NicMcPhee:
That's why we have code reviews! 😜

kklamberty:
> I'm not sure this is really great testing, but it does increase the coverage.

I have more confidence in the tests that are included in this PR now than I did before. I think it's worth adding these tests to the repository.

NicMcPhee (approved):
This looks great. We probably want to also test the `hashCode()` function, but that could be in a different issue.

PR #1301, opened 2023-02-23
Fix various layout and style issues

floogulinc:
Cleans up and fixes various layout and style issues left from the transition to MDC

### DIFF
--- a/client/src/app/app.component.scss
+++ b/client/src/app/app.component.scss
@@ -23,9 +23,9 @@
 
 .main-content {
   margin: auto;
-  margin-top: 24px;
+  padding: 24px;
   @media (max-width: 599px) {
-    margin: 16px;
+    padding: 16px;
   }
 
   // adds padding to the main content container on larger screens (above 1280px wide)
--- a/client/src/app/app.component.scss
+++ b/client/src/app/app.component.scss
@@ -22,8 +22,12 @@
 
 
 .main-content {
-  margin: 24px;
+  margin: auto;
+  margin-top: 24px;
   @media (max-width: 599px) {
     margin: 16px;
   }
+
+  // adds padding to the main content container on larger screens (above 1280px wide)
+  max-width: 1280px;
 }
--- a/client/src/app/home/home.component.html
+++ b/client/src/app/home/home.component.html
@@ -1,7 +1,14 @@
 <div class="flex-row">
   <div class="flex-1" fxFlex.gt-sm="80" fxFlexOffset.gt-sm="10">
-    <mat-card appearance="outlined" class="home-card">
-      This is a home page! It doesn't do anything!
+    <mat-card class="home-card">
+      <mat-card-header>
+        <mat-card-title>
+          Home
+        </mat-card-title>
+      </mat-card-header>
+      <mat-card-content>
+        This is a home page! It doesn't do anything!
+      </mat-card-content>
     </mat-card>
   </div>
 </div>
--- a/client/src/app/home/home.component.html
+++ b/client/src/app/home/home.component.html
@@ -1,5 +1,5 @@
 <div class="flex-row">
-  <div class="flex-1" fxFlex.gt-sm="80" fxFlexOffset.gt-sm="10">
+  <div class="flex-1">
     <mat-card class="home-card">
       <mat-card-header>
         <mat-card-title>
--- a/client/src/app/users/add-user.component.html
+++ b/client/src/app/users/add-user.component.html
@@ -11,11 +11,8 @@
               <mat-label>Name</mat-label>
               <input matInput placeholder="Name" formControlName="name"
                     required>
-              <mat-error *ngFor="let validation of addUserValidationMessages.name">
-                <mat-error class="error-message" data-test="nameError"
-                          *ngIf="addUserForm.get('name').hasError(validation.type) && (addUserForm.get('name').dirty || addUserForm.get('name').touched)">
-                  {{validation.message}}
-                </mat-error>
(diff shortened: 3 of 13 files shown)


client/src/app/app.component.scss
```diff
@@ -22,8 +22,12 @@
...
-  margin: 24px;
+  margin: auto;
+  padding: 24px;
   @media (max-width: 599px) {
-    margin: 16px;
+    padding: 16px;
   }
+
+  // adds padding to the main content container on larger screens (above 1280px wide)
+  max-width: 1280px;
```
kklamberty, line 32: Is 1280 the usual standard? It looks okay, but I do wonder if there are use cases where we might want to have content for the full width. (Probably not this one, but something more information intensive maybe.)

client/src/app/home/home.component.html
```diff
@@ -1,7 +1,14 @@
 <div class="flex-row">
-  <div class="flex-1" fxFlex.gt-sm="80" fxFlexOffset.gt-sm="10">
```
kklamberty, line 2: The `fxFlex` things were clearly not doing real work here, but I do wonder: did you put something in that would address what they used to address? Or, do we just not worry about that?

client/src/app/users/add-user.component.html
```diff
@@ -1,66 +1,68 @@
...
+            <mat-form-field>
+              <mat-label>Role</mat-label>
+              <mat-select required formControlName="role">
+                <mat-option value="admin">Admin</mat-option>
+                <mat-option value="editor">Editor</mat-option>
+                <mat-option value="viewer">Viewer</mat-option>
+              </mat-select>
+              <mat-error *ngIf="addUserForm.get('role').hasError('required')">You must make a selection</mat-error>
+              <mat-error *ngIf="addUserForm.get('role').hasError('pattern') && !addUserForm.get('role').hasError('required')">
+                Your selection is invalid
```
jpwalbran, line 62: Do we want a more specific error message here? I know this isn't a new change, but it does seem a little bit vague.

client/src/app/users/user-list.component.scss
```diff
@@ -5,6 +5,7 @@
 
 .user-card {
   margin-bottom: 10px;
+  flex: 1 1 280px;
```
kklamberty, line 8: I think documenting (for other developers and for the students) what this is doing would be helpful.

kklamberty (changes requested):
I am wondering what we could do for the add-user.component.html to make it so that errors in validation show up right by the field they are warning about. Right now, the spacing is very odd and it's somewhat hidden in some cases. I don't know why that's happening.

floogulinc:
> I am wondering what we could do for the add-user.component.html to make it so that errors in validation show up right by the field they are warning about. Right now, the spacing is very odd and it's somewhat hidden in some cases. I don't know why that's happening.

I'll have to take a look at that. I probably didn't test it and see. The solution is likely just spacing them out a bit more.

client/src/app/home/home.component.html
floogulinc: Yeah the new `max-width` for the main container does something similar but not the same, limiting the width on larger screens.

client/src/app/app.component.scss
floogulinc: It's one of the Angular Material breakpoints but which one is somewhat arbitrary. I suppose I could do something that better emulates what was behing done before and do a `max-width` of 80% on > small screens.

kklamberty: As long as I don't get wide margins when I have my window adjusted smaller just because my physical screen is large, I can live with that. I dislike when I am forced to have large buffers at the edge on ways that prohibit me from making my own choices about how to arrange my windows. 🙂

kklamberty:
@floogulinc - You can go ahead and revert my commit. I also failed to test things, and of course the tests are not passing - there were errors expected that I replaced with hints. My apologies.

floogulinc:
I've updated the add user form to use typed forms and also it now only shows a single error message at a time for each control (which is how the Angular Material form fields are designed). Which message is shown when there is multiple errors is based on the order the messages are listed in the `addUserValidationMessages` object.

kklamberty (dismissed):

kklamberty:
I think you're going to need to bring back the empty `ngOnInit` in AddUser

floogulinc:
> I think you're going to need to bring back the empty `ngOnInit` in AddUser

Why?

floogulinc:
> I think you're going to need to bring back the empty `ngOnInit` in AddUser

I'm not sure why `ngOnInit` was being manually called in the test but I have removed that and it all works.

kklamberty:
This looks good! Thanks for making these updates. Very much appreciated.

I might like it a bit better if the heights of cards in the same rows were the same. Right now, for folks with long names or company names, their cards can tend to make things look a little odd. 

I'm also unsure if there are ways to make the number of cards in the last row not determine the width of the cards. For example, if there are three columns in all the rows except the last row, I would prefer if the only card on that row still had a normal width instead of filling the whole width with one card. There may be cases where I would feel differently, but for the kind of data we have it feels a little unnatural to have some of them take on such different dimensions. 

Both of those minor things still look a bit better to me than the way things looked just removing the grid-ness of the layout like it was before your changes and after taking out flex-layout.

This is a nice set of improvements and brings us into the world of typed forms, too, which seems like a good thing!

kklamberty (approved):

floogulinc:
Ah I didn't test the user list all that much since I don't have seeding the db working on my end. I'll take a look at improving that if I get the chance. CSS grid might be the way to go.

PR #1302, opened 2023-02-23
Use more reactive techniques in components

floogulinc:


NicMcPhee:
What's your goal here? I was thinking about what might be something similar? Watching/reading stuff on the Internet, there seems to be an opinion out there that one should avoid calling `subscribe()` as much as possible, as that breaks up RXJS information flows. There seem to be arguments in favor of actually never subscribing in most cases, and then having async pipes in the HTML that realize the final `Observable`. I've been thinking of look at our code to see how feasible that would be, but never got there, so I was interested to see this PR.

floogulinc:
That's basically what I was working on. After the work we did on Buzz About we learned a lot about using rxjs optimally in Angular. I left this as a draft though after realizing that the tests would all need modification to work directly with observables and that wouldn't be done before the iteration started.

floogulinc:
I pretty much have a good idea of the road forward on this it's just a question of putting in the time and whether that's even worth doing if we might be replacing Angular which is something we've discussed a little.

NicMcPhee:
Could we do an example or two of better using RXJS now? That could benefit @kklamberty's current students, perhaps. That might be useful even if we never end up making those changes across the entire project.

floogulinc:
Well the changes I made to the user profile component in this PR already could be but the tests still need to all be updated.

NicMcPhee:
This may have now been made at least partially moot by the move towards signals. I think we could remove most of our use of RxJS with signals, keeping RxJS for just the services that have to do async things like HTTP requests.

PR #1304, opened 2023-02-23
Replace mongo iterables with into()

### LINKED ISSUE #1303: Restructure Mongo tests in Mongospec.java
I noticed that the way the tests handled aggregation in Mongospec.java was unnecessarily redundant. The methods store the result of an aggregation pipeline in an AgregationIterable<> opbject, and then pipes them into a list, which is one step more than necessary, and it isn't that much of an extra cognitive load to combine those two steps (which is what the docs do for the aggregation pipeline anyway)

jpwalbran:
Replaced the temporary variable assignments using the AgregateIterable() and FindIterable types when it was immediately cast into an ArrayList, instead just coercing them directly into ArrayLists.
Closes #1303

### DIFF
--- a/server/src/test/java/umm3601/mongotest/MongoSpec.java
+++ b/server/src/test/java/umm3601/mongotest/MongoSpec.java
@@ -140,7 +140,7 @@ class MongoSpec {
   }
 
   @Test
-  public void over25SortedByName() {
+  void over25SortedByName() {
     List<Document> docs
       = userDocuments.find(gt("age", 25))
       .sort(Sorts.ascending("name"))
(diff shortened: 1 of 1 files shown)


kklamberty (changes requested):
Was there a reason to switch the access level to public?  That was a code smell that I just addressed as Deep Source suggested, and if we don't need these to be public for any reason, I would suggest changing them to package-level like they were.

jpwalbran:
Nice catch! That was an accidental holdover from a quick merge conflict resolution, and they should definitely be package level instead.

PR #1305, opened 2023-02-25
Make match lab 4

kklamberty:
The main goal of this pull request is to bring elements from the most recent lab into the iteration template

### DIFF
--- a/client/src/app/users/user-profile.component.html
+++ b/client/src/app/users/user-profile.component.html
@@ -1,5 +1,6 @@
 <div class="flex-row">
-  <div class="flex-1">
+  <div class="flex-1" *ngIf="user; else loading">
     <app-user-card [user]="this.user"></app-user-card>
   </div>
 </div>
+<ng-template #loading>Loading User Profile data...</ng-template>
--- a/client/src/app/users/user.service.spec.ts
+++ b/client/src/app/users/user.service.spec.ts
@@ -228,7 +228,7 @@ describe('UserService', () => {
           .toHaveBeenCalledTimes(1);
         expect(mockedMethod)
           .withContext('talks to the correct endpoint')
-          .toHaveBeenCalledWith(userService.userUrl + '/' + targetId);
+          .toHaveBeenCalledWith(`${userService.userUrl}/${targetId}`);
       });
     }));
   });
@@ -295,7 +295,7 @@ describe('UserService', () => {
       // paying attention to what is returned (undefined) didn't work well here,
       // but I'm putting something in here to remind us to look into that
       userService.addUser(testUsers[1]).subscribe((returnedString) => {
-        console.log('The thing returned was:' + returnedString);
+        console.log(`The thing returned was: ${returnedString}`);
         expect(mockedMethod)
           .withContext('one call')
           .toHaveBeenCalledTimes(1);
--- a/client/src/app/users/user.service.spec.ts
+++ b/client/src/app/users/user.service.spec.ts
@@ -1,10 +1,11 @@
-import { HttpClient } from '@angular/common/http';
+import { HttpClient, HttpParams } from '@angular/common/http';
 import { HttpClientTestingModule, HttpTestingController } from '@angular/common/http/testing';
-import { TestBed } from '@angular/core/testing';
+import { TestBed, waitForAsync } from '@angular/core/testing';
+import { of } from 'rxjs';
 import { User } from './user';
 import { UserService } from './user.service';
 
-describe('User service: ', () => {
+describe('UserService', () => {
   // A small collection of test users
   const testUsers: User[] = [
     {
@@ -47,10 +48,10 @@ describe('User service: ', () => {
     TestBed.configureTestingModule({
       imports: [HttpClientTestingModule]
     });
-    httpClient = TestBed.inject(HttpClient);
-    httpTestingController = TestBed.inject(HttpTestingController);
     // Construct an instance of the service with the mock
     // HTTP client.
+    httpClient = TestBed.inject(HttpClient);
+    httpTestingController = TestBed.inject(HttpTestingController);
     userService = new UserService(httpClient);
   });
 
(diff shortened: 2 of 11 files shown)


client/src/app/users/user-profile.component.html
```diff
@@ -1,5 +1,6 @@
 <div class="flex-row">
-  <div class="flex-1">
+  <div class="flex-1" *ngIf="user; else loading">
```
jpwalbran, line 2: Thanks for finally adding this!

client/src/app/users/user.service.spec.ts
```diff
@@ -59,135 +60,249 @@ describe('User service: ', () => {
...
-
-    req.flush({id: 'testid'});
+  describe('Adding a user using `addUser()`', () => {
+    it('talks to the right endpoint and is called once', waitForAsync(() => {
+      // Mock the `httpClient.addUser()` method, so that instead of making an HTTP request,
+      // it just returns our test data.
+      const USER_ID = 'pat_id';
+      const mockedMethod = spyOn(httpClient, 'post').and.returnValue(of(USER_ID));
+
+      // paying attention to what is returned (undefined) didn't work well here,
```
jpwalbran, line 295: Are you saying that the post request is expected to return undefined? That shouldn't be the case. I don't know whether this is really blocking or not, but it seems weird to get an undefined anywhere here.

jpwalbran (commented):
Mostly looks good, there's just one thing that probably needs some discussion.

client/src/app/users/user.service.spec.ts
```diff
@@ -59,135 +60,249 @@ describe('User service: ', () => {
...
+
+    it('correctly calls api/users with filter parameter \'admin\'', () => {
+        const mockedMethod = spyOn(httpClient, 'get').and.returnValue(of(testUsers));
+
+        userService.getUsers({ role: 'admin' }).subscribe((users: User[]) => {
+          // The array of `User`s returned by `getUsers()` should be
+          // the array `testUsers`. This is "weird" because we'd truly be expecting
+          // the server to return just `admin` users, but as mentioned above, we're
+          // not trying to get the server here.
+          expect(users)
```
NicMcPhee, line 122: Given the comments up above, do we even want to bother checking what `users` gets returned by `getUsers()`? Maybe all we really care about is that it gets called once (line 125) and with the right params (line 132).

The only argument I can see for testing this is to make sure that `getUsers()` does actually return what it gets back from the server, and doesn't modify that list in some way. If that's a reason to keep it, then we'd want to change the comment above this line to say that.

Thoughts?

client/src/app/users/user.service.spec.ts
```diff
@@ -59,135 +60,249 @@ describe('User service: ', () => {
...
 
-    // Check that the role, company, and age parameters are correct
-    expect(req.request.params.get('role')).toEqual('editor');
-    expect(req.request.params.get('company')).toEqual('IBM');
-    expect(req.request.params.get('age')).toEqual('37');
+    it('correctly calls api/users with filter parameter \'age\'', () => {
+      const mockedMethod = spyOn(httpClient, 'get').and.returnValue(of(testUsers));
+
+      userService.getUsers({ age: 25 }).subscribe((users: User[]) => {
+        expect(users)
```
NicMcPhee, line 142: Drop this as well?

client/src/app/users/user.service.spec.ts
```diff
@@ -59,135 +60,249 @@ describe('User service: ', () => {
...
+          //      * And a "age:37" pair.
+
+          // This gets the arguments for the first (and in this case only) call to the `mockMethod`.
+          const [url, options] = mockedMethod.calls.argsFor(0);
+          // Gets the `HttpParams` from the options part of the call.
+          // `options.param` can return any of a broad number of types;
+          // it is in fact an instance of `HttpParams`, and I need to use
+          // that fact, so I'm casting it (the `as HttpParams` bit).
+          const calledHttpParams: HttpParams = (options.params) as HttpParams;
+          expect(users)
```
NicMcPhee, line 175: And drop this?

client/src/app/users/user.service.spec.ts
```diff
@@ -59,135 +60,249 @@ describe('User service: ', () => {
...
+      // We have to `subscribe()` to the `Observable` returned by `getUserById()`.
+      // The `user` argument in the function below is the thing of type User returned by
+      // the call to `getUserById()`.
+      userService.getUserById(targetId).subscribe((user: User) => {
+        // The `User` returned by `getUserById()` should be targetUser.
+        // This `expect` doesn't do a _whole_ lot.
+        // This really just confirms that `getUserById()`
+        // doesn't in some way modify the user it
+        // gets back from the server.
+        expect(user)
```
NicMcPhee, line 223: The comment here is the kind of thing that we'd probably want on the `expect(users)` lines up above if we keep them. If we don't keep them, then maybe we want to get rid of this `expect` as well.

client/src/app/users/user.service.spec.ts
NicMcPhee: Yeah, I don't think you shouldn't have gotten `undefined` back here. If you did, then perhaps the mocking wasn't complete? I don't immediately see the problem, though, but it does seem like something's wrong.

NicMcPhee (changes requested):

server/src/test/java/umm3601/user/UserControllerSpec.java
```diff
@@ -210,6 +216,34 @@ void canGetUsersWithAge37() throws IOException {
     assertEquals(2, userArrayListCaptor.getValue().size());
   }
 
+  @Test
+  void canGetUsersWithAge37Redux() throws JsonMappingException, JsonProcessingException {
```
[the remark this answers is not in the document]
kklamberty: I am inclined to either include both and make it clear why they are both there or bring the argument matcher in to the one that doesn't have it and talk about the different ways of checking. I disagree that it would likely lead to students doing more testing than is needed. I don't think they use the presence of a test as a cue... they look at their coverage and what they need to test. I agree that removing the extra test classes we had was a good move, but I think multiple examples could serve them well.

client/src/app/users/user.service.spec.ts
```diff
@@ -59,135 +60,249 @@ describe('User service: ', () => {
...
-    const req = httpTestingController.expectOne(
-      (request) => request.url.startsWith(userService.userUrl)
-        && request.params.has('role') && request.params.has('company') && request.params.has('age')
-    );
-
-    // Check that the request made to that URL was a GET request.
-    expect(req.request.method).toEqual('GET');
+  describe('When getUsers() is called with parameters, it correctly forms the HTTP request (Javalin/Server filtering)', () => {
+    /*
+    * We really don't care what `getUsers()` returns in the cases
```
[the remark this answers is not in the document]
kklamberty: Thanks! I was trying to help clarify in yet another place for the students the differences between the client and server filtering.

client/src/app/users/user.service.spec.ts
kklamberty: Just to clarify (because it's oddly hard for me to track which things are "the comments above" and be sure I understand what's going on...):

- There are comments in the code about how we are checking that we got back all of the users since the filtering would be taken care of by the server.
- Since we only actually care about talking to the right endpoint and how many times things were called, we could just remove the parts that verify that we got all the users back.
- So... for anything that is only filtered on the server, we could remove the part that is checking something we actually don't care about? (e.g., age and role are only filtered on the server)

client/src/app/users/user.service.spec.ts
kklamberty: Do we drop the three lines so this `expect` goes away? (And, to be clear, it goes away since are is filtered on the server anyway.)

client/src/app/users/user.service.spec.ts
kklamberty: Here I am less confident... does this line and it's related `withContext` and `toEqual` go away? I think it does since this is a call to `getUsers` and we get all the users back in this test. Is that correct?

client/src/app/users/user.service.spec.ts
kklamberty: I think we should leave this comment and leave this expect where it is. Does removing it clarify anything?

client/src/app/users/user.service.spec.ts
kklamberty: I've now created an issue to explore this further.

kklamberty: Thinking back to when I first wrote this comment (maybe for lab 3?), I think it was not something that was undefined in the test or mock, but in the console it was undefined. I think I'll need to have some help figuring out what we actually want to do here.

server/src/test/java/umm3601/user/UserControllerSpec.java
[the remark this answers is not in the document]
NicMcPhee: I've seen _numerous_ occasions where students copy "unnecessary" tests through all the way to their final projects, even when they're labelled as "just examples". In the heat of battle there's a lot of copy/paste, and when there's limited understanding folks tend to just copy and tweak.

I do agree that a variety of examples can be useful, though, and I don't want to discourage that. I just don't want to do it in a way that ends up cluttering up the code down the road. Maybe if we have multiple versions in Lab 4 but _not_ in the iteration template that would be better? I'm not sure.

If we do include both, then we definitely want to comment the heck out of them both and make it really clear that they're effectively redundant and that people should _not_ in general have multiple tests of the same thing like this.

I'm also fine with copying the argument matcher into the one we keep if you think that would be useful.

Maybe we should go through all the tests and try to identify key features like argument matchers that we want to illustrate somewhere. Then we could actually list the features, along with which test(s) highlight/use that feature. That might provide some structure and help students see these more as a resource to learn from and less as a pile of copy to mechanically copy and tweak?

client/src/app/users/user.service.spec.ts
NicMcPhee: No, I think it definitely is tricky to make sense of context when you add a comment in a review. All that shows up here is the link you marked and a quite small number of lines above/around it, and that makes it hard to know what things refer to. I'll try to do a better job of copying in more info in comments like this in future.

Your summary is pretty much what I was thinking. We have a bunch of tests where we return, e.g., all the the users, and then check that we return all the users, even though we might reasonably expect some kind of filtering to have happened. Given that in the test we say what the HTTP request should return, it's not clear what the value is of testing that we get it back. As it says somewhere in the comments, about the only reason to check that is to make sure that `getUsers()` or whatever doesn't mess with or manipulate those values.

So I guess the "big" question is whether we want to include those checks or not. I'm really not sure what I think would be the best approach, but it seems that we want to be consistent whatever we choose to do. My (weak) inclination is to leave out those checks since they are mostly just checking "obvious" things, but I can be convinced otherwise.

client/src/app/users/user.service.spec.ts
NicMcPhee: It's the `expect` that I'm suggesting we drop (unless we want to keep the tests that check that `getUsers()` returns what the HTTP request returns.

client/src/app/users/user.service.spec.ts
NicMcPhee: Again, it's the `expect` that I'm suggesting we might drop.

client/src/app/users/user.service.spec.ts
```diff
@@ -59,135 +60,237 @@ describe('User service: ', () => {
...
+      //
+      // We have to `subscribe()` to the `Observable` returned by `getUsers()`.
+      // The `users` argument in the function is the array of Users returned by
+      // the call to `getUsers()`.
+      userService.getUsers().subscribe((users: User[]) => {
+        // The array of `User`s returned by `getUsers()` should be
+        // the array `testUsers`.
+        expect(users)
+          .withContext('expected users')
+          .toEqual(testUsers);
```
NicMcPhee, line 80: These are the three lines that we _might_ want to remove if we wanted to remove all the tests that "just" check that `getUsers()` doesn't mess with the list of users returned by the `HttpClient`.

```suggestion
```

client/src/app/users/user.service.spec.ts
```diff
@@ -59,135 +60,237 @@ describe('User service: ', () => {
...
+    * about the returned value).
+    */
+
+    it('correctly calls api/users with filter parameter \'admin\'', () => {
+        const mockedMethod = spyOn(httpClient, 'get').and.returnValue(of(testUsers));
+
+        userService.getUsers({ role: 'admin' }).subscribe(() => {
+          expect(mockedMethod)
+            .withContext('one call')
+            .toHaveBeenCalledTimes(1);
```
NicMcPhee, line 122:
```suggestion
```

And here.

client/src/app/users/user.service.spec.ts
```diff
@@ -59,135 +60,237 @@ describe('User service: ', () => {
...
+      // the call to `getUserById()`.
+      userService.getUserById(targetId).subscribe((user: User) => {
+        // The `User` returned by `getUserById()` should be targetUser.
+        // This `expect` doesn't do a _whole_ lot.
+        // This really just confirms that `getUserById()`
+        // doesn't in some way modify the user it
+        // gets back from the server.
+        expect(user)
+          .withContext('expected user')
+          .toBe(targetUser);
```
NicMcPhee, line 213:
```suggestion
```

This would be another equality check that we may want to remove if we decide we don't want to keep these tests.

client/src/app/users/user.service.spec.ts
```diff
@@ -59,135 +60,237 @@ describe('User service: ', () => {
...
+      //
+      // We have to `subscribe()` to the `Observable` returned by `getUserById()`.
+      // The `user` argument in the function below is the thing of type User returned by
+      // the call to `getUserById()`.
+      userService.getUserById(targetId).subscribe((user: User) => {
+        // The `User` returned by `getUserById()` should be targetUser.
+        // This `expect` doesn't do a _whole_ lot.
+        // This really just confirms that `getUserById()`
+        // doesn't in some way modify the user it
+        // gets back from the server.
```
NicMcPhee, line 210:
```
        // This `expect` doesn't do a _whole_ lot.
        // This really just confirms that `getUserById()`
        // doesn't in some way modify the user it
        // gets back from the server.
```

This is the kind of comment that we should keep and repeat in all the places where we have these "unnecessary" tests.

NicMcPhee (changes requested):

server/src/test/java/umm3601/user/UserControllerSpec.java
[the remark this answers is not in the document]
kklamberty: I'm inclined to make an issue about this and still go ahead and include them both in the code for this year. I really want to get this out the door, and I am willing to deal with whatever comes our way this semester as a result of including both. I agree that pointing out great elements of all of the tests and being really careful about what to include would be wise, but I also think that including it as it is and then having conversations about it will be fine for this year. I don't want this to hold up the whole PR, and I'm not in a position to fix everything this semester. I've tackled quite a few issues and made things a lot more consistent between the iteration template and the labs than it was. I'm not going to do the major overhaul this term, even though it might be a good thing.

client/src/app/users/user.service.spec.ts
kklamberty: I didn't remove it from here, but this one might also be good to remove, and the comments that are on the calls that *do* have parameters should probably be here too (I would say not instead, though, because it was more obvious in all the other places.)

client/src/app/users/user.service.spec.ts
kklamberty: Good point

NicMcPhee (approved):

PR #1309, opened 2023-02-27
Beef up the karma test of "misbehaving" user list

kklamberty:
The `UserListComponent` can't populate the UserList if the server gives it an error rather than giving it useful data. There are a variety of reasons this might happen... one of which is that the server is down. There was an existing test that only said that the UserList would be undefined if there was not a service available. However, there were not any checks that were specific about what happened in that case. This PR adds to what we expect in that case.


kklamberty:
Since we are not saying what the mocked service should return as far as exactly what kind of error, I think this is all we can really test here. Perhaps we should be saying more about what the error message should be, including the error code, but I think this actually works okay as it is: the user list is encountering the right branch of the code based on what we would expect if the server was not giving the client an updated user list (because of an error). It updates the message, but since we didn't say what code, that part is undefined.

client/src/app/users/user-list.component.ts
```diff
@@ -32,6 +32,7 @@ export class UserListComponent implements OnInit, OnDestroy  {
   public userCompany: string;
   public viewType: 'card' | 'list' = 'card';
 
+  errMsg = '';
```
NicMcPhee, line 35: Is there a reason you moved this out of `getUsersFromServer()`? It's only used there, so it seems that it would make more sense as a local variable there. Is it necessary to ensure that the message passes through to the snackbar in some way?

Oh, I think I get it now. You needed it to be at the class level so you could access it in the tests. Is that right? If so, we might want to document that.

client/src/app/users/user-list.component.spec.ts
```diff
@@ -146,10 +142,26 @@ describe('Misbehaving User List', () => {
   }));
 
   it('generates an error if we don\'t set up a UserListService', () => {
+    const mockedMethod = spyOn(userList, 'getUsersFromServer').and.callThrough();
     // Since calling either getUsers() or getUsersFiltered() return
     // Observables that then throw exceptions, we don't expect the component
     // to be able to get a list of users, and serverFilteredUsers should
     // be undefined.
-    expect(userList.serverFilteredUsers).toBeUndefined();
+    expect(userList.serverFilteredUsers)
```
NicMcPhee, line 150: This is really nice – thanks!

client/src/app/users/user-list.component.spec.ts
```diff
@@ -146,10 +142,26 @@ describe('Misbehaving User List', () => {
...
+      .withContext('service can\'t give values to the list if it\'s not there')
+      .toBeUndefined();
+    //expect(fixture.detectChanges)
```
NicMcPhee, line 153: Why is this code commented out. Should we remove it?

NicMcPhee (changes requested):
This looks really nice. I had a few small comments, but I definitely think we should merge this in soon.

The one "bigger" thing I wanted to ask is whether we want to have any tests that handle client side errors. Are there "common" client side errors that we can fake and test? I'm honestly not sure what kinds of errors are likely/plausible on the client side here, so I'm not sure how much sense this makes. We shouldn't let this block the PR, though; we can also make a new issue for that so we can merge this in.

kklamberty:
> Note that the branch coverage is still only 50% since when we updated the error handling, we put both server and client errors, and we've really only tested the server errors. We should either remove the other case (if it's not needed here) or test the other case (but I'm not yet sure what that would look like).

Nic - you (rightly pointed out that the test is commented out. Here was my earlier comment about that. I can't think of anything in particular to test, so I agree that that may need to be a different issue. It's been almost a year, so I don't remember too clearly what I was thinking, so I can only go with what I wrote (which was that I was not sure what that would look like).

client/src/app/users/user-list.component.spec.ts
kklamberty: Is this a change request? The review says you requested changes, but I'm not seeing where those are. I'm happy to have you delete (or to delete myself) this unused code if that would do it.

NicMcPhee: Yeah, I was proposing that we remove that commented out code. I think that having commented out code (especially without some note explaining why it's there and commented out) is just confusing for the students. And it then gives the impression that it's OK for them to have loads of commented out code in their projects, which we don't want.

So if we remove this then I think this is good to go!

NicMcPhee:
Oops – I think we both did the same thing at the same time. 😜

NicMcPhee (approved):

PR #1316, opened 2023-03-02
Filter controller component

kklamberty:
I've found it a bit tricky to describe what the user-list component is doing and to separate out the display of the users from the filtering of the users. This PR focuses on adding a new component, the user-filter, that shows the controls for filtering and takes care of providing a collection of users that have been filtered in the requested way. 

Adding even one more component really made the list of components in the users folder quite long. I decided to also try moving away from a flat structure to a structure that puts components' files into related folders so that it was a little more browsable and manageable. I think it also might help make components stand out from the user-service or other non-components. 

Thoughts?


client/src/app/users/user-list/user-list.component.html
```diff
@@ -5,70 +5,28 @@
...
-              <mat-option>--</mat-option>
-              <mat-option value="admin">Admin</mat-option>
-              <mat-option value="editor">Editor</mat-option>
-              <mat-option value="viewer">Viewer</mat-option>
-            </mat-select>
-            <mat-hint>Filtered on server</mat-hint>
-          </mat-form-field>
-        </div>
-
-        <br>
```
kklamberty, line 50: This is mostly moved to the `user-filter.component.html`

client/src/app/users/user-list/user-list.component.html
```diff
@@ -5,70 +5,28 @@
       <mat-card-header>
         <mat-card-title class="user-list-title">Users</mat-card-title>
       </mat-card-header>
+      <app-user-filter [filteredUsers]="filteredUsers" (filteredUsersChange)="filteredUsers=$event"></app-user-filter>
```
kklamberty, line 8: This tells what bindings to use and what events to pay attention to.

client/src/app/users/user-filter/user-filter.component.ts
```diff
@@ -1,40 +1,34 @@
...
-  public filteredUsers: User[];
-
-  public userName: string;
-  public userAge: number;
-  public userRole: UserRole;
-  public userCompany: string;
-  public viewType: 'card' | 'list' = 'card';
+export class UserFilterComponent implements OnInit, OnDestroy {
+  @Output() readonly filteredUsersChange = new EventEmitter<User[]>();
+  @Input() filteredUsers: User[];
```
kklamberty, line 23: Making these into input and output and stating where to emit changes (line 93) keeps the filteredUsers up-to-date.

client/src/app/users/user-filter/user-filter.component.ts
```diff
@@ -89,21 +87,10 @@ export class UserListComponent implements OnInit, OnDestroy  {
...
-  }
-
-  /**
-   * Starts an asynchronous operation to update the users list
-   *
-   */
-  ngOnInit(): void {
-    this.getUsersFromServer();
+      this.serverFilteredUsers, { name: this.targetUserName, company: this.targetUserCompany });
+    this.filteredUsersChange.emit(this.filteredUsers);
```
kklamberty, line 93: Let the world know that I have updated my filteredUsers.

NicMcPhee:
I haven't look at the code yet, but the motivation in your description sounds awesome. I really like the idea of separating the filtering from the display, and I think that having a "richer" directory structure sounds like that could be a big help. I've seen a lot of projects where I think students really should split things up in their organization, but they often stick to a very flat system because it's what they're "used to".

NicMcPhee (approved):
Very cool – thanks for making this valuable change!

client/src/app/users/user-filter/user-filter.component.html
```diff
@@ -0,0 +1,44 @@
+
```
floogulinc, line 1: Why is this component a partial portion of a card and not the whole card?

floogulinc (changes requested):
I have some concerns about this PR. I don't think the component with the filter inputs should be responsible for actually requesting users from the server as is being done here. That should happen in the higher level user list component. Using an output from a sub component to provide data from a service is not really the right flow I think. You generally want data to always cascade down by being passed into inputs rather than sending it up through outputs (outputs should be reserved for when the sub component is a control of some type and needs to tell the parent component it's state). 

I don't think splitting the filter inputs and the list is really needed but If you want to I would suggest having the component with the filter inputs just output the current filter settings which would be used by the list component when getting the users from the service.

I actually think it would be much nicer to split the portion that displays the list out into its own component but that' still seems excessive considering how small the user list component already was.

As for the reorganization into folders for each component, that is good but I think we should just use the structure the Angular CLI makes by default where within `src/app` there is a folder per component (eg remove the extra `users` folder).

client/src/app/users/user-filter/user-filter.component.ts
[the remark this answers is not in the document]
kklamberty: I was reading about two-way binding, and this was part of that example. I think (if I recall correctly... it's been a few days now) if I removed it, things were unhappy. It sounded like it was important for the event emitter to have the same name as the input except for the word "change" being there.

client/src/app/users/user-filter/user-filter.component.html
kklamberty: I suppose it could be different. The only reason I did it that way was because I was not aiming to change the look of the previous page, but I was wanting to pull the filtering out of the list.

kklamberty:
> I have some concerns about this PR. I don't think the component with the filter inputs should be responsible for actually requesting users from the server as is being done here. That should happen in the higher level user list component. Using an output from a sub component to provide data from a service is not really the right flow I think. You generally want data to always cascade down by being passed into inputs rather than sending it up through outputs (outputs should be reserved for when the sub component is a control of some type and needs to tell the parent component it's state).
> 
Interesting. Could you link me to some particular place where the direction of the flow and the relationship between components and services is described? I was thinking that the user-list component including all these filter controls was a bit odd since we then also say in the "list" that you can view it as a "list" or a "card". I was imagining that we kind of had a view of the users that was either a grid of user cards or a list of users, and both of those are getting information triggered by these filters. BUT, you could imagine that those views of collections of users could be associated with a variety of different filter controls. One is a list view. One is a grid view. Either of those needs a collection of users to display, but neither really needs to have anything to do with the filtering. I thought it worked nicely to say that either one would need an input of an array of users and could display those users appropriately.

> I don't think splitting the filter inputs and the list is really needed but If you want to I would suggest having the component with the filter inputs just output the current filter settings which would be used by the list component when getting the users from the service.
> 
Can you say more about that? I don't understand why we would want the list to also take care of doing the filtering or being part of the querying process. The filters are about filtering, and the lists are about listing.

> I actually think it would be much nicer to split the portion that displays the list out into its own component but that' still seems excessive considering how small the user list component already was.
> 
I disagree that it was small, and I think I did split the portion that _displays the list_ out into its own component. In fact, what you describe sounds less like it's keeping the list separate. Can you elaborate? I'm quite confused.

> As for the reorganization into folders for each component, that is good but I think we should just use the structure the Angular CLI makes by default where within `src/app` there is a folder per component (eg remove the extra `users` folder).

client/src/app/users/user-filter/user-filter.component.ts
kklamberty: I'm not sure if this was the case when you posted your comment, but we use `@Input` in UserCardComponent. One question that @floogulinc asked in UserListComponent is why use an `@Input` there since the UserListComponent is not embedded anywhere. This is in contrast with the UserCardComponent, which is used in the card view of the list and in the profile view of a single user. So, maybe this is an important distinction I didn't know about. Perhaps I was imposing a parent/child relationship between the filter component and the list component?

kklamberty:
I think we should split this into two parts and make new issues to address them (with their own PRs if they get picked up as important). 

#1455 Organize Angular Components Into Folders
#1456 Split filter controls out of the user list component

PR #1317, opened 2023-03-02
Make package private rather than public

kklamberty:
This PR addresses a couple of minor problems with the level of access stated for a test class and a test method that Deep Source pointed out.


NicMcPhee (approved):
This looks great. I didn't realize that JUnit 5 now allows package private classes and test methods, and that the recommendation is to use package private as the most restrictive allowed approach (thus providing the strongest encapsulation).

PR #1355, opened 2023-11-30
Update Gradle wrapper to latest version of Gradle (v8.5)

NicMcPhee:
This should update Gradle to v8.5, which is the latest version as of December 2023.

In doing this I also added a Gradle provisioning plugin that (I think) will install missing JDKs as necessary. I think we had thought a previous update had set up Gradle so that it would download missing JDKs, but that never actually worked. Now I think it actually does, and you can run, e.g., `./gradlew build` on a machine that doesn't have Java 11 (the current required version) and it will download and install it for you. This should be a big help if students want to set all this up on their own computers.


kklamberty (approved):
This looks good to me. I was able to run the the server and all the tests pass for me as well.

PR #1366, opened 2023-12-15
Improve the `UserController` test coverage

NicMcPhee:
This gets the `UserController` test coverage up to 100% by using `Mockito.spy` to create mock user controllers that we can use to test the `generateAvatar` function.

These tests are based on code suggested by GitHub CoPilot, although I had to make a few minor changes to get them to pass. The CoPilot code used

```java
        UserController userController = Mockito.spy(UserController.class);
```

to generate a mock controller. This failed, I think because not enough was mocked out. I changed it to instead be:

```java
    UserController controller = Mockito.spy(userController);
    when(controller.md5(email)).thenReturn("md5hash");

    String avatar = controller.generateAvatar(email);

    assertEquals("https://gravatar.com/avatar/md5hash?d=identicon", avatar);
```

and

```java
    UserController controller = Mockito.spy(userController);
    when(controller.md5(email)).thenThrow(NoSuchAlgorithmException.class);

    String avatar = controller.generateAvatar(email);

    assertEquals("https://gravatar.com/avatar/?d=mp", avatar);
```

Here we mocked the `userController` field constructed in the `setupEach()` method. This provided an actual controller that could be called, and only used `.spy()` to overwrite the `.md5()` method.

These tests pass and we get 100% coverage of the server code!


kklamberty (approved):
This is a thing of beauty! So nice to have that last part tested! Thanks for doing this.

PR #1369, opened 2023-12-16
Improve add user component coverage

### LINKED ISSUE #1307: Figure out more about the karma test about adding a user in `UserServiceSpec`
Yeah, I don't think you shouldn't have gotten `undefined` back here. If you did, then perhaps the mocking wasn't complete? I don't immediately see the problem, though, but it does seem like something's wrong.

_Originally posted by @NicMcPhee in https://github.com/UMM-CSci-3601/3601-iteration-template/pull/1305#discussion_r1118011143_

I posted a comment in the code that was noticed in code review about what happens with the addUser test. I seemed to get back `undefined` rather than something useful. We should look into that. Here is the test:
```typescript
  describe('Adding a user using `addUser()`', () => {
    it('talks to the right endpoint and is called once', waitForAsync(() => {
      // Mock the `httpClient.addUser()` method, so that instead of making an HTTP request,
      // it just returns our test data.
      const USER_ID = 'pat_id';
      const mockedMethod = spyOn(httpClient, 'post').and.returnValue(of(USER_ID));

      // paying attention to what is returned (undefined) didn't work well here,
      // but I'm putting something in here to remind us to look into that
      userService.addUser(testUsers[1]).subscribe((returnedString) => {
        console.log('The thing returned was:' + returnedString);
        expect(mockedMethod)
          .withContext('one call')
          .toHaveBeenCalledTimes(1);
        expect(mockedMethod)
          .withContext('talks to the correct endpoint')
          .toHaveBeenCalledWith(userService.userUrl, testUsers[1]);
      });
    }));
  });
```

NicMcPhee:
This uses some GitHub CoPilot suggests to get the test coverage of `AddUserComponent` up to 100%.

There were two methods that weren't tested: `getErrorMessage()` and `submitForm()`, and this adds tests for both of those.

You could argue that some of this (especially the testing for `getErrorMessage()`) is a bit fragile since it's tied to specific error messages. If that concerns folks we can work on trying to make it less fragile.

---

As almost an accident, I think I also fixed #1307 along the way.

The previous test here didn't work because we were expecting to get a `string` ID back, when the HTTP request actually returns a _JSON object_ back, which contains an `id` field that has the `string` ID.

This is arguably a type error, and one might hope that the TypeScript type checker would have helped us here. Unfortunately the Mockito `.returnValue()` method can take any value so the type system couldn't tell that we were passing a `string` when a JSON object was expected.

If @kklamberty disagrees, she should remove #1307 from the list of issues this will close (over on the right).


client/src/app/users/user.service.spec.ts
```diff
@@ -297,15 +297,17 @@ describe('UserService', () => {
 
   describe('Adding a user using `addUser()`', () => {
     it('talks to the right endpoint and is called once', waitForAsync(() => {
+      const user_id = 'pat_id';
```
NicMcPhee, line 300: The previous test here didn't work because we were expecting to get a `string` ID back, when the HTTP request actually returns a _JSON object_ back, which contains an `id` field that has the `string` ID.

This is arguably a type error, and one might hope that the TypeScript type checker would have helped us here. Unfortunately the Mockito `.returnValue()` method can take any value so the type system couldn't tell that we were passing a `string` when a JSON object was expected.

kklamberty (approved):
This looks great! Good catch on the String versus JSON object... Tricky to notice!

PR #1389, opened 2023-12-23
Remove JDK8 dependencies

NicMcPhee:
We had some Java/Gradle dependencies that I think were there to support Java v8, which we really shouldn't be using anymore. I think this removes those and causes the system to use more modern versions of the relevant Java libraries.

I tested these (remotely) in the lab and everything appears to work fine. We're currently using Java v11 in the lab, and we definitely want to move to something higher since Java is now in the v20+ realm. Definitely no need to still care about v8.


willster000 (commented):
According to some quick googling, Jackson Databind allows you to convert Java objects into JSON format and vice versa. Is that functionality no longer needed? Or are you achieving it in some other way?

NicMcPhee:
Good question. I was honestly a little surprised that I could remove that and not have things break.

[This part of the Javalin docs](https://javalin.io/documentation#configuring-the-json-mapper) suggests that Javalin brings in key parts of Jackson on its own, so I think it's actually doing the data binding (probably via Jackson) under the covers, which may be why I could remove it and not blow up the world.

I just searched through our code to see where/how we used Jackson directly, and it seems like there are two uses:

- `JacksonMongoCollection` in the `UserCollection` (which actually comes from the `mongojack` dependency, and
- `JavalinJackson` in the test code (which comes from Javalin)

So at least in the iteration template we're not using Jackson stuff directly. It might be worth looking at some previous projects and make sure _they_ don't use Jackson in unexpected ways, though.

NicMcPhee:
@willster000 – I had a quick look at Iteration 3 of one of the S23 projects, and it looks like all the uses of Jackson in that codebase are essentially the same as in the iteration template. So I don't think we need to explicitly depend on `databind` anymore (if we ever did).

kklamberty (approved):
This looks for to me. I am able to run all of the tests and run the server on my machine. I approve. Thanks @willster000 for the good question and @NicMcPhee for the follow up on that. I think it will be safe to update.

Issue #675, opened 2022-01-31
Add tests for the `Server` class


NicMcPhee:
There's a Javalin testing tutorial that demonstrated using a library called Unirest to fake the HTTP requests and responses that we'd need to test the `Server` class directly. If we added these tests we'd

- Have better end-to-end coverage of the server code
- Be able to include `Server` in the test coverage calculations

kklamberty:
It's not totally clear to me if we'd want to use the "built in" Javalin Test or Unirest to do this work, but I wanted to be sure we connect these two issues/PR to each other since they seem related.

NicMcPhee:
Thanks for connecting those dots, @kklamberty. I no longer remember if there's a meaningful difference between what I was trying in [that PR](https://github.com/UMM-CSci-3601/3601-lab2_client-server/pull/224) and this one. I made them both at about the same time, but in different repos, so who knows what I was thinking.

I might try a version with Unirest to see if it helps the port problem. Even if it does, though, we have the potential issue that @floogulinc raised of this wiping the test database. I'm a little less worried about that, since presumably the test database is there to be wiped out and rebuilt. So as long as we're not running these tests at the same time as some other tests, we should be good.

I also added an issue (#1391) that raises the question of whether we should replace actual Mongo calls in _unit tests_ with mocking/spies. I'm not sure if we should do that, but I wanted to raise the issue.

PR #1397, opened 2023-12-29
Refactor `Server#main`

NicMcPhee:
This is a fairly major refactoring of the rather long `main()` method in the `Server` class. I _think_ this is on balance a Good Thing, but it would be good to get feedback from others before we merge it.

The key new methods are (in roughly order of most significant to least):

- `setupRoutes` encapsulates setting up the routes.
   - This is probably the best thing here IMO. It currently doesn't do much, but it provides a natural place for students to put _their_ routing when the time comes.
   - It also moves the actual routing to the `UserController` class so it's part of that class instead of out in the "general" `Server` class. 
- `configureDatabase()` encapsulates the database configuration.
- `configureJavalin()` encapsulates the configuration of the instance of the Javalin server.
   - `configureShutdowns()` (called by `configureJavalin()` handles the somewhat complex shutdown logic.
- `startServer()` sets up and starts the Javaline server
- `getEnvOrDefault()` (slightly) simplifies the handling of environment variables.

This may provide some opportunities to test some of these extracted methods (initially there was no testing for this class), but I think I'd rather that happened in one or more separate PRs.

---

I realized as I did this that we could use interfaces to reduce the coupling between `Server` and the controllers (in our case just the `UserController`:

- Create a `Controller` interface with (so far) just a `setupRoute()` method.
- Pass an array of `Controller`s to the `Server` constructor. Then the `Server` code can be independent of the specifics of any particular controller. That would also make it harder to for students to insert (possibly untested) logic into the `Server` class since everything would be constrained by the interface.
- I had to make a defensive copy of the `controllers` array in the `Server` constructor since DeepSource (correctly) considered that a major risk. Kind of annoying, TBH, but the way Java's type system handles arrays leaves us without better options. (In Rust we could pass ownership of the array, and its contents, to the server and we wouldn't have worry about this.)

That arguably should be a separate PR, but it's pretty entangled in these changes so I included it here. If that is troublesome for folks, I can pull it out into its own PR.

---

This whole set of changes once again bumps up against the annoying problem of testing Javaline routes. This has been previously discussed in an issue in this repository (#675: "Add tests for Server class") and from a draft PR for Lab 2 ([Add Javalin functional tests](https://github.com/UMM-CSci-3601/3601-lab2_client-server/pull/224)). The problem is that the `Server` class has _zero_ test coverage, and is in fact removed from the test coverage calculations by this part of `build.gradle`:

```gradle
  afterEvaluate {
    classDirectories.setFrom(files(classDirectories.files.collect {
      fileTree(dir: it, exclude: 'umm3601/Server.class')
    }))
  }
```

Sometimes students have (probably innocently) ended up sticking a lot of logic in `Server`, where it wasn't tested and that wasn't flagged by the coverage calculations.

I'm hoping that this refactoring (and especially refactoring out the routing) might make this better in some ways. I'm not sure we want to put a lot of effort into writing tests for `Server`, since a lot of that is just going to be confirming that things don't break since we don't want to be testing Mongo's implementation of their database or Javalin's implementation of their server. The big potential win is that we can continue to exclude `Server` from the coverage calculations, but I think students are now less likely to put much code here. Hopefully they'll follow the pattern in `setupRoutes()` and put their routes (and any other logic) in their controllers.

One downside of this is that the route configuration can't be easily tested, as discussed in the issue and PR linked above. Now we're pushing the route configuration into their controllers, so there's going to be a bit of their code that they can't easily test. We could exclude the `UserController#setupRoutes()` method from the coverage calculations and give them permission to exclude a similar method from their controllers, although that could be abused. It's a short method with no branching or complex logic, so not having it covered wouldn't hurt their coverage stats much, so maybe it's OK. (The coverage for `UserController` in this branch is currently 93%, and it'll go up one or two percentage points when we merge in #1366.)

I had hoped that I'd be able to test the route setup by spying on `server` and ensuring that methods like `.get()` and `.post()` were called with the right paths and functions to call. Sadly, it turns out that you can't confirm that the function that's passed in (e.g., `userController::addNewUser`) is correct because every use of `userController::addNewUser` appears to generate a (different) lambda. So if I try to use a line like:

```java
    Mockito.verify(server).post(UserController.API_USERS, userController::addNewUser);
```

to check that something like this is done in the route setup:

```java
    server.post(API_USERS, this::addNewUser);
```

it doesn't work because each of those `::addNewUser`s is apparently allowed to (and does on my computer with the current version of Java) create a new lambda which the system sees as different. Thus the `Mockito.verify()` call fails because the two lambdas are aren't the same. Reading the Internet some makes me think there's nothing that we can really do about that :cry:, but I'd be happy to be proved wrong on that.

Many of the refactoring suggestions came from asking GitHub CoPilot for refactoring suggestions. I didn't take all its suggestions, and I modified some (sometimes necessary to get them to work). There are, however, a number of changes here suggested by CoPilot that I'd never thought of. The idea of refactoring out `setupRoutes()` and moving the specific user route setup to `UserController`, for example, came from CoPilot and is IMO a nice change.


willster000:
I'm a little confused by having both the new interface Controller and the old class UserController which implements the Controller interface. What's the relationship between these two entities? 

If I'm understanding correctly, the server calls the addRoutes method for all controllers at the bottom of the file. Let's say a student has added a route for a get request to the Controller.java file which already exists in UserController.java. What happens when that route is called?

NicMcPhee:
Good questions @willster000. The `Controller` interface says what methods any controller will _provide_ which, in this case (so far) is just `addRoutes()`. Then _any_ controller will need to implement that interface and provide an `addRoutes()` method.

When the students implement their project, they'll probably want a controller for each primary type in the project. If it was a shopping app, for example, there would probably be a `ProductController`, and a `ShoppingCartController`, and an `OrderController`, etc., etc. Each one of those will need to implement the `Controller` interface, providing an `addRoutes` method.

The nifty thing, then, is that the `Server` code is almost completely separate from the specifics of those controllers. All it knows is that it gets an array of `Controller`, without knowing what _kind_ of controllers those are. It does know that it can call `addRoutes()` on each of them since that's specified in the `Controller` interface.

This might be clearer if we pulled `main()` out into a new `Main` class, which then constructs the `Server` and starts it. Then `Main#main` would know about `UserController` and `ProductController`, but `Server.java` would have zero references to any concrete controllers.

Actually, I think I really like that idea and will try to add that feature if I can find time later. That way we can say that `Main` should be ignored in the test coverage (there's no good way to test the `main()` method of a program without something like E2E tests), but `Server` will still be covered.

Does this all make some sense? Interfaces are really powerful, but can be confusing, and I know we don't do a _ton_ with them in Data Structures or Software Design these days, so this may all seems a bit new.

NicMcPhee:
I went ahead and created `Main.java` and moved `main()` and a couple of helper methods into that. This now means that `Server.java` has _no_ dependencies on any of the specific controllers (e.g., `UserController`).

A question for the students (e.g., @willster000): Do you think it would be easier to understand for students if this example code had two or more controllers? We expect students to make multiple controllers, but we only have one in our example. Does that make it harder for students to see the pattern(s) and create the expected structure?

Alternatively, should we have a `Server structure` Markdown document that explains the server structure (including the `Controller` interface)? Would students read something like that, or (as I suspect) mostly just look at and try to follow patterns in the code?

willster000:
Your answer about the controller interface is very helpful. Thanks! This may just be a me-problem but I was misinterpreting the comment in the interface to mean that you should implement specific routes in the controller, which was my main confusion. Do think this comment ought to be changed?

public interface Controller {
  /**
   * Add routes for the server.
   *
   * This is where you'll add all the routes for your data, by calling
   * `server.get(...)`, `server.post(...)`, etc.
   *
   * @param server The Javalin server to add routes to
   */
  void addRoutes(Javalin server);
}

To your question, I think two controllers would help students see patterns in the controllers (probably to the point where a server markdown wouldn't be necessary) but in order for the second controller to do something, which perhaps isn't necessary at all, we would need to add a whole lot to the front end... correct? If we do add another example controller that doesn't have any corresponding components in front end, I think we should clearly state that it is an example. Otherwise I could see students pattern-recognizing at a higher level than they should and being utterly confused (e.g. "I thought controllers had something to do with processing requests from the client but this controller with no corresponding front end seems to be a counter example to that")

NicMcPhee:
Good point about the comment. I'll try to make that clearer.

I agree that if we add a second controller on the backend, there really ought to be a corresponding front-end component. That's part of why this has never happened. (There's been discussion about adding a little more to the example, but we've never actually done it.)

That said, there's no reason we couldn't try to do it now. I could imagine a simple change that provides a company-based views, where there's some additional info about the company, for example. Then there would be a full stack of code, from display on the Angular end, through a controller on the server end, and a new collection in the DB.

server/src/main/java/umm3601/Controller.java
```diff
@@ -0,0 +1,15 @@
+package umm3601;
+
+import io.javalin.Javalin;
+
+public interface Controller {
+  /**
+   * Add routes for the server.
```
NicMcPhee, line 7: @willster000 correctly observed that this comment makes it sound like you actually add the routes in the interface _which isn't true_. We should modify this to make it clearer that this is "done" and any additional work happens in classes that implement this interface.

NicMcPhee (commented):
@willster000 : If you had made a comment like this in a review of the code (clicking "Request changes"), then it would have essentially become an to-do on the PR. I can't actually click "Request changes" (or "Approve") on a self-review, so it's just a comment when I do it.

willster000 (changes requested):
Hopefully this creates a to-do within the PR like Nic mentioned under my comment about the comment in the controller interface.

NicMcPhee:
In the most recent commit ([f5cc777](https://github.com/UMM-CSci-3601/3601-iteration-template/pull/1397/commits/f5cc777cbfb6451b3492b34b50191221ad494bc6)), I expanded the documentation of the `Controller` interface 

In the process, I actually renamed the `Controller` interface to be the `AddsRoutes` interface since that is really what this interface is about. It doesn't really describe whole controllers, it just describes that specific aspect of controllers.

While I think this is in some ways the "right" thing to do, it's possible it might make things more confusing for the students given their limited experience with interfaces. If people thing it would be better to change the name back to `Controller` I can totally do that.

kklamberty (commented):
I don't think that using interfaces will be a problem in general. We make use of interfaces pretty frequently in data structures. However, I think the use of a name that describes a behavior might be confusing. What about something like ControllerRouteAdder? Or, is it a ControllerWithRoutes? I think it needs to be a noun. The methods can have verb-y names.

server/src/main/java/umm3601/Main.java
```diff
@@ -42,16 +43,21 @@ static String getEnvOrDefault(String envName, String defaultValue) {
...
+   * Get the implementations of `AddsRoutes` used for the server.
+   *
+   * These will presumably be one or more controllers, each of which
+   * implements the `AddsRoutes` interface. You'll add your own controllers
+   * in `getRouteAdders` as you create them.
    *
    * @param database The MongoDB database object used by the controllers
    *               to access the database.
-   * @return An array of controllers for the server.
+   * @return An array of implementations of `AddRoutes` for the server.
```
kklamberty, line 54: I think the use of "AddRoutes" here is a bit confusing. This will be an array of controllers that implement the interface you renamed... ("AddsRoutes"... which is mentioned here as "AddRoutes" and in "getRouteAdders"). I feel like there is a little inconsistency that should be fixed. line 27 kind of implies it will be RouteAdders.

server/src/main/java/umm3601/Main.java
```diff
@@ -42,16 +43,21 @@ static String getEnvOrDefault(String envName, String defaultValue) {
...
    */
-  static Controller[] getControllers(MongoDatabase database) {
-    Controller[] controllers = new Controller[] {
-      // You would add additional controllers here, as you create them.
+  static AddsRoutes[] getRouteAdders(MongoDatabase database) {
+    AddsRoutes[] controllers = new AddsRoutes[] {
+      // You would add additional controllers here, as you create them,
+      // although you need to make sure that your new controllers implement
```
kklamberty, line 59: maybe change to "although you need to make sure that each of your new controllers implements"

kklamberty (changes requested):
I think you might need to revert back to calling these Controller and still include your extra documentation that helps remind people about how interfaces work. The change of name is a bit confusing since the words are not always the same or in the same order, sometimes being more like an adjective or property (AddsRoutes) and other times more like a verb (AddRoutes) or a noun (RouteAdder).

I'm not sure if we might also be losing some convention for servers by not referring to these as controllers (especially if we want them to call things UserController or WhateverController).

I think change it back to Controller (but keep some of your other documentation) or change it to be more consistently noun, verb, or adjective/property (and I think it's mostly best to avoid the verb).

NicMcPhee:
I do agree that the new naming does complicate things in ways that probably aren't super helpful. My goal (probably not consistently achieved) was

- To have the interface be an adjective, as it describes things that implement it.
- To have concrete implementations be nouns, as they are the things that implement the interface.
- To have the method be a verb, as it's an action that we take.

An an example in another context, we might have an `Edible` interface that specifies an `eat()` method. Then things like `Scone` and `Burger` and `Carrot` could all implement `Edible`.

If I thought there would be other interfaces introduced into the server code, then I might want to fight for more focused names. But I don't know why there would be, TBH, and we probably don't want to turn this into some kind of lesson on interface design. I'll try to make those changes and ask for for another review.

Thanks!

server/src/main/java/umm3601/AddsRoutes.java
```diff
@@ -0,0 +1,38 @@
...
+ * Any class that implements this interface can be used to add routes to the
+ * server via the specified `addRoutes()` method.
+ *
+ * This is useful for organizing routes into separate files, and for testing
+ * routes without starting the server (except that the inability to compare
+ * lambdas in fact makes this very hard to test).
+ *
+ * Any "controller" class that provides routes for the Javalin server
+ * must implement this interface since the `Server` class
+ * is just handled an array of `Controller` objects in its constructor. This
```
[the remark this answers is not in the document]
NicMcPhee: Thanks – fixed.

server/src/main/java/umm3601/Main.java
NicMcPhee: Good catch – thanks.

NicMcPhee:
Commit [e2c576d](https://github.com/UMM-CSci-3601/3601-iteration-template/pull/1397/commits/e2c576d219e6674bcc7c817cc3b2f01d8f11a319) renames the `AddRoutes` interface back to `Controller` and (hopefully) updates all the comments and variable/function names to reflect that change. Maybe this is ready to merge now?

server/build.gradle
```diff
@@ -79,12 +79,12 @@ jacocoTestReport {
...
   afterEvaluate {
-    // This excludes the `Server` class from the coverage report. We don't
-    // have any good way to test the `Server` class directory (we'd have
+    // This excludes the `Main` class from the coverage report. We don't
+    // have any good way to test the `Main` class (we'd have
     // to somehow fake incoming HTTP requests), so we are just
     // leaving it out of the coverage report and the coverage limits.
     classDirectories.setFrom(files(classDirectories.files.collect {
-      fileTree(dir: it, exclude: 'umm3601/Server.class')
+      fileTree(dir: it, exclude: ['umm3601/Main.class', 'umm3601/Server.class'])
```
kklamberty, line 87: The words say we will exclude Main, but we are also excluding Server. Is that intentional? If so, should both be used/listed in the words?

server/src/main/java/umm3601/Server.java
```diff
@@ -6,73 +6,120 @@
...
+   *
+   * This "wires up" the database using either system environment variables
+   * or default values. If you're running the server locally without any environment
+   * variables set, this will connect to the `dev` database running on your computer
+   * (`localhost`). If you're running the server on Digital Ocean using our setup
+   * script, this will connect to the production database running on server.
+   *
+   * This sets both the `mongoClient` and `database` fields
+   * so they can be used when setting up the Javalin server.
+ * @param mongoAddr The address of the MongoDB server
```
kklamberty, line 59: For some reason, this is aligned differently from the other lines. Maybe it's just missing a space?

server/src/main/java/umm3601/Server.java
```diff
@@ -86,5 +133,45 @@ public static void main(String[] args) {
...
+    });
+  }
+
+  /**
+   * Setup routes for the server.
+   *
+   * @param server The Javalin server instance
+   */
+  private void setupRoutes(Javalin server) {
+    // Add the routes for all the implementations of `Controller` in the
```
kklamberty, line 171: I'd use "each of" rather than "all" since it's a for each loop.

kklamberty (dismissed):
I really like this refactoring of the Server. I think you are correct that it will lead to a more thoughtful treatment of controllers and putting routes in the controllers instead of changing the server code. Thank you for all of your work to make this refactoring make sense!

I made a couple of minor comments that you can either address or now. I don't think they are crucial.

server/build.gradle
NicMcPhee: I had to keep `Server` excluded because without the PRs that added all the `Server` tests, the code coverage wasn't high enough to pass. Now that you've approved those PRs, let me look at the code coverage and see what happens.

If it doesn't get us there, my inclination is to merge this in since it changes so much, and then create an issue to de-exclude `Server`, adding tests as necessary.

server/src/main/java/umm3601/Server.java
NicMcPhee: No idea what happened there. This is why we should add `prettier` to the project (which there's a semi-old issue for).

I'd rather not block this merge over that if possible, but it obviously does need to get fixed. 😢

server/src/main/java/umm3601/Server.java
NicMcPhee: Yup – that would be better. I'm not in a position to make these changes right now. We're really up against the deadlines at this point, tomorrow won't be any better, and Wednesday we fly. Can you make them and re-review the code? (Will it let you re-review is you pushed the most recent changes?) If not, I'll try to get back to this when I get home.

kklamberty (approved):

NicMcPhee:
Thanks to @kklamberty for wrapping this up!

PR #1407, opened 2024-01-08
Add CodeQL support

NicMcPhee:
The auto-generated PR (#1242) is now out of date with the world, and I can't update that because it belongs to them. So I created a new (up-to-date) branch, merged their changes in, and we'll see if that works. If it does, then we should merge this in and close #1242.

This just adds one file, so reversing it should be easy if we don't like what it did.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,42 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+  schedule:
+    - cron: "25 11 * * 0"
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ javascript, java ]
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v2
+        with:
+          languages: ${{ matrix.language }}
+          queries: +security-and-quality
+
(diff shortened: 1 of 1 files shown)


NicMcPhee:
This seems to have doubled the number of checks, and not in a helpful way. I think that I had turned on these checks through a previous change, and that this turns them on again in a different way. That makes me think we may not want to merge this in.

NicMcPhee:
OK, I think I understand what's going on now. There are two ways to configure CodeQL:

- The "simple" way via the Settings interface
- The "advanced" way by committing a YAML file with the config info in it

This PR (and #1242) take the YAML file approach.

I had independently set up CodeQL via the Settings interface because I noticed it while futzing around and decided it would be interesting to turn it on.

We can't have both.

I think the only difference between this advanced (YAML) configuration and what I did is which test suite is being used to check the code. I just used the _default_ suite, and this YAML config says to use the _extended_ suite. I just changed the simple config to also use the extended suite, so they should be functionally the same now.

At first I was thinking we'd just close this PR and stick to using the simple configuration. It occurred to me, though, that we probably want the YAML configuration, because then it will be part of the repositories and will be picked up and enabled by student forks of this repository. If it's in the WebUI configs, then it _won't_ propagate to student forks, and they won't have these quality checks (which I'm assuming we'd want them to have?).

Thoughts on this @kklamberty?

kklamberty:
Hmmmm. I think that the YAML idea (for the reasons you state) makes the most sense. Will people *also* need to turn on some kind of option, or is the YAML entirely *instead of* needing to even flip some switch?

NicMcPhee:
I'm not _100%_ sure, but my hope is that if we go the YAML route they'll magically pick that up and it'll run those checks without them having to do anything else. If you approve this PR and we merge it in, one of us can then fork the repo and see what happens.

kklamberty (approved):

PR #1432, opened 2024-01-14
1417 switch from ngmodule to standalone 

### LINKED ISSUE #1417: Switch from `NgModule` to standalone components
`NgModule` has always been really confusing for students (and me!), and thanks to the introduction of standalone components, I think we don't need `NgModule` any more.

So I think we want to set all our components to be:

```typescript
@Component({
  standalone: true
  ...
})
```

and add the necessary component imports for anything the component depends on:

```typescript
@Component({
  ...
  imports: [OurOtherComponent, SomeOtherThing, YetAnotherThing]
```

I also think we need to modify `main.ts` to call something like:

```typescript
bootstrapApplication(AppComponent, appConfig)
    .catch((err) => console.error(err));
```

I _think_ this can also really simplify our unit testing, because we won't need all the module stuff there.

This is a place where it would almost be easier to re-build the Angular app from scratch since then it would magically not have any of the `NgModule` stuff in it from the start. Hopefully we can just refactor to this universe, though.

NicMcPhee:
This follows the instructions on Angular's [Migrate and existing Angular project to standalone](https://angular.io/guide/standalone-migration) documentation, using their `ng generate @angular/core:standalone` tool to do all the heavy lifting.

The first of the three steps did most of the work. The second step didn't do anything (so there's no commit for it), and the third commit didn't have a lot to do.

Closes #1417


kklamberty (approved):
That makes a lot more imports, but it does, indeed, seem a bit less magical and mysterious. It was easy to forget to do something with the modules, and this probably helps with that. Thanks for working through these changes!

Issue #139, opened 2020-07-18
Improve error handling/reporting on unknown users


NicMcPhee:
Currently, if the Angular app requests an unknown user, or the server returns something completely inappropriate, the app just quietly swallows the problem instead of doing some kind of reasonable error handling.

There are two related problems, both of which should be addressed:

1. If the client requests an unknown ID (if, e.g., I add `xyz` to the end of every ID), then there's a message on the console, but basically a blank screen in the browser. We should catch and handle the `HttpResponseError`; [this doc page](https://angular.io/guide/http#handling-request-errors) will probably be helpful.
2. If the server returns something of the wrong type (I had it return a new `HashSet` instead of a `User`, for example) then the app just quietly swallows the problem and don't report anything _anywhere_. Nothing is on the console, and the app displays a user profile with blank strings for everything.

For (1) we should presumably display an alert or dialog letting the user know that a bad request was made. This is fundamentally a software error instead of a user error, so they should probably be asked to notify the developers? By adding `xyz` to the IDs, I got a 400 `BadRequestError`; if you changed it so that it was a legal Mongo ID that wasn't in the DB I suspect you'd get a 404 `NotFoundError` instead. We'd want to catch any/all of these.

For (2) we should presumably the server did Bad Things and that the user should again report the errors to the developers.

In both cases it would be nice to provide enough info in the dialog that the user would have something fairly concrete to provide the developers.

We also ought to have tests that handle all these cases. I wouldn't be surprised to find that there's a hole in the test coverage around this somewhere.

This is related to #53 in some ways; both issues will certainly end up touching the same code.

NicMcPhee:
This is related to #348

NicMcPhee:
PR #1442 handles part of this (basically the first item in the list above), but doesn't do anything about the second one.

PR #1442, opened 2024-01-14
Improve error handling when failing to load a user profile

### LINKED ISSUE #348: User Profile error state
When the user profile fails to load the user it should show _something_ rather than just a blank page.

NicMcPhee:
At some point in the past, if the app failed to load a user profile, the resulting user profile page was completely blank, as described in #348 and #139. Somewhere along the way we addressed parts of that, but the error message wasn't at all helpful, and having it in a disappearing snackbar didn't seem great.

In my Twitch stream today, we worked on improving this, and there is now the following error card on the user profile page if the user fails to load:

image

The "title" message is in this code. The next line ("The request user id …") comes from the Javalin server and is the error message provided in the `UserController`. The last line ("Http failure…") is the actually HTTP response error message.

I also completely removed the snackbar since it didn't seem at all necessary now.

Closes #348 

Handles part of #139, but doesn't do all of it.


client/src/app/users/user-profile.component.scss
```diff
@@ -0,0 +1,6 @@
+.error-card {
+  background-color: #b00020;
```
NicMcPhee, line 2: I don't like having this color hard-coded here. I tried using the color palette info from the themes but couldn't get it to work. This is the hex code from the red provided in the error palette, though.

client/src/app/users/user-profile.component.ts
```diff
@@ -13,6 +13,7 @@ import { map, switchMap, takeUntil } from 'rxjs/operators';
 })
 export class UserProfileComponent implements OnInit, OnDestroy {
   user: User;
+  error: { help: string, httpResponse: string, message: string };
```
NicMcPhee, line 16: We added this new `error` field to `UserProfileComponent` that holds the (currently three) parts of the error message. Those are then used in the HTML to display the different parts of the message.

There is an argument for having this be a proper, named class instead of an anonymous object like this, so feel free to push back on this.

client/src/app/users/user-profile.component.ts
```diff
@@ -54,9 +55,11 @@ export class UserProfileComponent implements OnInit, OnDestroy {
         return user;
       },
       error: _err => {
-        this.snackBar.open('Problem loading the user – try again', 'OK', {
-          duration: 5000,
-        });
+        this.error = {
+          help: 'There was a problem loading the user – try again.',
+          httpResponse: _err.message,
+          message: _err.error ? _err.error.title : '',
```
NicMcPhee, line 61: It took me quite a while to figure out how to get at the message we added in the Javalin code. It turns out it was in the `title` field in the `error` object. In the end I figured that out by (at the suggestions of esitsu@Twitch) printing `JSON.stringify(_err.error)`, which showed me to the structure of that object and its fields. Nothing I had searched gave me any indication where that message was, so using `JSON.stringify()` was a real lifesaver.

client/src/app/users/user-profile.component.scss
kklamberty: This seems like something @floogulinc might know about. I would love to get a recorded walkthrough about all of the theming stuff so that I could understand that better.

client/src/app/users/user-profile.component.ts
kklamberty: I don't have an immediate, gut reaction about using a proper, named class or not. Might that make the other error handling stuff easier?

client/src/app/users/user-profile.component.ts
kklamberty: Is it a bad thing that we are not looking at an error code? Or, did I just miss something and you do have an error code in there? I see that the error code 400 is displayed. Is the title where that comes from and it's not a special part of the error?

kklamberty (approved):
This addition looks good, but it does reduce our test coverage since there are no added tests for it. I'm not opposed to having you merge it in, but maybe tests would be a good thing to add?

client/src/app/users/user-profile.component.scss
NicMcPhee: Me too! I don't feel like I've ever understood what @floogulinc did there, and I'm not sure the students have been able to use the theming in a very effective way.

client/src/app/users/user-profile.component.ts
NicMcPhee: Maybe? It depends on whether this sort of structure is re-usable in other settings.

Maybe that's the way to deal with this – leave it alone for now, and see if we end up creating the same or something similar elsewhere. If we do, extract that out into a named class. If we don't, then leave it alone.

client/src/app/users/user-profile.component.ts
NicMcPhee: Actually, the "400" that is displayed is in the `_err.message`, which is the HTTP response message the client receives.

The `_err.error` object also contains a `status` field that is `"400"`. So we could have a test that looks at `_err.error.status` and provides different messages depending on the status. I actually thought about doing that, but that sounds complicated to do _well_ and I'd rather this PR not get bogged down in that given how close we are to the start of the semester. I made an issue (#1451) for it, though, and we can bump up its priority if you think it's important.

NicMcPhee:
Thanks for catching the test coverage change – I hadn't noticed that. I'll look into it and report back.

kklamberty (approved):
Looks good and works for me!

kklamberty (approved):

PR #1453, opened 2024-01-15
1452 Deal with large bundle size issues

### LINKED ISSUE #1452: Deal with large bundle size issues
Somewhere (I think in upgrading to v17 of Angular?) our bundle size started exceeding 5MB, which generates an error when you run `ng build`.

```text
Warning: bundle initial exceeded maximum budget. Budget 2.00 MB was not met by 3.13 MB with a total of 5.13 MB.

Error: bundle initial exceeded maximum budget. Budget 5.00 MB was not met by 135.49 kB with a total of 5.13 MB.
```

That doesn't seem good, but I'm also not sure what we can do about it. It looks like the vast majority of the bundle size is in Angular (`vendor` in the output of `ng build` below):

```text
Initial Chunk Files           | Names         |  Raw Size
vendor.49690c607c5a3793.js    | vendor        |   4.78 MB | 
styles.977a00c33a08a58c.css   | styles        | 163.21 kB | 
polyfills.49ed01ba5725107e.js | polyfills     | 104.06 kB | 
main.ac1934fb7c2ac4e6.js      | main          |  89.93 kB | 
runtime.b9d774f8a405e829.js   | runtime       |   5.91 kB | 
```

Our code is just under 90KB, so that's hardly a major issue.

Despite being listed as an `Error`, this doesn't seem to break our builds, but it doesn't look Good and seems like we should try to figure out how to make things better.

kklamberty:
When working on a different PR, we noticed that our bundle size started exceeding the budget (#1452). 

It occurred to me that it might be due to changes involved with upgrading to Angular 17. There is some information about a new builder that this PR integrates that significantly reduces the size of our bundle. [Using the application builder](https://angular.io/guide/esbuild#using-the-application-builder) significantly reduces the size of our bundle.

There may be other things to consider that are not addressed here, but this PR gets us below the error level. I also changed the budget for the initial bundle to be 4mb so we would not get a warning every time we build. We may want to do some other optimizations to fully address the issue, but I think that this PR closes #1452.

### DIFF
--- a/client/angular.json
+++ b/client/angular.json
@@ -50,7 +50,7 @@
               "budgets": [
                 {
                   "type": "initial",
-                  "maximumWarning": "2mb",
+                  "maximumWarning": "4mb",
                   "maximumError": "5mb"
                 },
                 {
--- a/client/angular.json
+++ b/client/angular.json
@@ -15,12 +15,14 @@
       "prefix": "app",
       "architect": {
         "build": {
-          "builder": "@angular-devkit/build-angular:browser",
+          "builder": "@angular-devkit/build-angular:application",
           "options": {
             "outputPath": "dist/client",
             "index": "src/index.html",
-            "main": "src/main.ts",
-            "polyfills": "zone.js",
+            "browser": "src/main.ts",
+            "polyfills": [
+              "zone.js"
+            ],
             "tsConfig": "tsconfig.app.json",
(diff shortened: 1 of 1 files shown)


NicMcPhee (approved):

NicMcPhee:
Very nice – thanks!

Issue #875, opened 2022-04-23
Introduce multi-line strings in the Java code


NicMcPhee:
The Java spec files have some really complicated string append stuff that's super hard to read, very error prone, and extremely difficult to debug. I think that a nice way to improve that code would be to use `String.format()` with multi-line strings. Those weren't supported until Java 15, so we'd need to switch to move to a _much_ newer version of Java. If we're going to bother updating, we might as well move to the most recent LTS, which is Java 17.

As example, in `UserControllerSpec.java` there's currently this:

```java
    String testNewUser = "{"
        + "\"name\": \"Test User\","
        + "\"age\": 25,"
        + "\"company\": \"testers\","
        + "\"email\": \"test@example.com\","
        + "\"role\": \"viewer\""
        + "}";
```

The simplest improvement would be to switch to multi-line strings like (which requires at least Java 15):

```java
    String testNewUser = """
      {
        "name": "Test User",
        "age": 25,
        "company": "testers",
        "email": "test@example.com",
        "role": "viewer"
      }
      """
```

If we get that working, then we can extract a function that will use `String.format()` to crank these things out. E.g.,

```java
    String makeTestUser(name, age, company, email, role) {
        return
            String.format(
             """
      {
        "name": "%s",
        "age": %s,
        "company": "%s",
        "email": "%s",
        "role": "%s"
      }
      """,
        name, age, company, email, role)
```

kklamberty:
Neat! For age, I think we'd want a number, but this otherwise looks really neat. Unfortunately, I think we're still in Java 11 land at the moment. I'm not sure if we want to ask about moving up, though, because that does look neat!

PR #1461, opened 2024-01-20
Replace long string appends with multi-line strings

NicMcPhee:
This replaces a bunch of unreadable and unmaintainable string append messes with single multi-line strings. These are much more readable and maintainable.

GitHub CoPilot was able to do most of this automatically, although I had to run it on each string separately. I suspect there's a prompt that would have had it change all of them in the file, but I didn't find it.

Fixes: #875

### DIFF
--- a/server/src/test/java/umm3601/user/UserControllerSpec.java
+++ b/server/src/test/java/umm3601/user/UserControllerSpec.java
@@ -427,13 +427,15 @@ class UserControllerSpec {
 
   @Test
   void addUser() throws IOException {
-    String testNewUser = "{"
-        + "\"name\": \"Test User\","
-        + "\"age\": 25,"
-        + "\"company\": \"testers\","
-        + "\"email\": \"test@example.com\","
-        + "\"role\": \"viewer\""
-        + "}";
+    String testNewUser = """
+        {
+          "name": "Test User",
+          "age": 25,
+          "company": "testers",
+          "email": "test@example.com",
+          "role": "viewer"
+        }
+        """;
     when(ctx.bodyValidator(User.class))
       .then(value -> new BodyValidator<User>(testNewUser, User.class, javalinJackson));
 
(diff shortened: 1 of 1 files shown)


server/src/test/java/umm3601/user/UserControllerSpec.java
```diff
@@ -480,13 +484,14 @@ void addInvalidEmailUser() throws IOException {
...
-        + "\"company\": \"testers\","
-        + "\"email\": \"test@example.com\","
-        + "\"role\": \"viewer\""
-        + "}";
+    String testNewUser = """
+        {
+          "name": "Test User",
+          "age": "notanumber",
+          "company": "testers",
+          "email": test@example.com",
```
kklamberty, line 492: I think a quote is missing before the given email here

kklamberty (changes requested):
I'm pretty sure there is a missing quote in there where I indicated, but this is *SO* much more readable than the previous version! Thanks for doing this!

server/src/test/java/umm3601/user/UserControllerSpec.java
NicMcPhee: You are correct! I'll fix that.

NicMcPhee: Thanks to @kklamberty for noticing that I was missing an open quote on the `"email"` field in the `addInvalidAgeUser()` test.

Weirdly, this totally passed the test because it _also_ generated a `ValidationExcption`. In doing some looking around, I couldn't find any easy way to figure out _where_ validation failed, so that we could distinguish between an intentional error like `"age": "notanumber"` and a type error like leaving off a quote.

That's a bummer because it's pretty easy to have a typo in this kind of "uninterpreted" JSON, and this means we would never know unless some eagle-eye like KK notices.

If someone knows of a better way to check for this, definitely let us know!

kklamberty (approved):

PR #1482, opened 2024-02-05
Fix control flow migration oddities

### LINKED ISSUE #1480: Fix @ control flow migration oddities
There are a few oddities that I think were the result of the automated update to the new control flow tools. These include:

- [x] We want to track `user._id`, not `user` in `@for` loops
- [x] There are two nested instances of `@if (this.simple)` in `UserCardComponent` HTML

NicMcPhee:
I've noticed a few odd bits that I think resulted from running the automated control flow migration tool. 

- [x] We want to track `user._id`, not `user` in `@for` loops
- [x] There are two nested instances of `@if (this.simple)` in `UserCardComponent` HTML

This fixes these issues.

Fixes #1480

### DIFF
--- a/client/src/app/users/user-card.component.html
+++ b/client/src/app/users/user-card.component.html
@@ -35,12 +35,10 @@
         </mat-list>
       </mat-card-content>
     }
-    @if (this.simple) {
-      <mat-card-actions>
-        @if (this.simple) {
-          <button mat-button data-test="viewProfileButton" [routerLink]="['/users', this.user._id]">View Profile</button>
-        }
-      </mat-card-actions>
-    }
+    <mat-card-actions>
+      @if (this.simple) {
+        <button mat-button data-test="viewProfileButton" [routerLink]="['/users', this.user._id]">View Profile</button>
+      }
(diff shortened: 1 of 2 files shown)


kklamberty (dismissed):
This looks reasonable to me. I was wondering if it mattered *which* of the repeated conditions to keep, but I guess that the tests passing is probably a good indicator.

NicMcPhee:
> This looks reasonable to me. I was wondering if it mattered _which_ of the repeated conditions to keep, but I guess that the tests passing is probably a good indicator.

I've actually tried it both ways, and they both seem to work. Looking at it again, though, I wonder if I kept the wrong one. If we kept the outer one then we wouldn't create an empty `mat-card-actions`, which seems like it might be better than what I have here. Should I change that?

kklamberty:
I think keeping the outer one sounds like a good plan if either one works.

NicMcPhee:
I've made that change – thanks for helping me see that!

kklamberty (approved):
Looks good! I think that was a sensible change.

PR #1484, opened 2024-02-06
Update a variety of `npm` package dependencies

NicMcPhee:
This brings everything up-to-date as of today.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -24,9 +24,9 @@
     "@angular/platform-browser": "^17.1.2",
     "@angular/platform-browser-dynamic": "^17.1.2",
     "@angular/router": "^17.1.2",
-    "rxjs": "~7.8.0",
+    "rxjs": "^7.8.0",
     "tslib": "^2.4.0",
-    "zone.js": "~0.14.3"
+    "zone.js": "^0.14.3"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


client/package.json
```diff
@@ -24,9 +24,9 @@
     "@angular/platform-browser": "^17.1.2",
     "@angular/platform-browser-dynamic": "^17.1.2",
     "@angular/router": "^17.1.2",
-    "rxjs": "~7.8.0",
+    "rxjs": "^7.8.0",
```
kklamberty, line 27: In lab 3, you have  "rxjs": "~7.8.1", but this is the only mismatch I noticed between lab 3 and here. I'm not sure what your intention is there.

kklamberty (commented):
I will approve this if you would like, but I wanted to draw your attention to a mismatch between lab 3 and this update in case it was an error.

client/package.json
NicMcPhee: Thanks for asking about this! I think the "right" answer is to use the `^` instead of `~`, so what we probably want to do is change Lab 3.

kklamberty: And, do you want the 7.8.1 or 7.8.0? Here, it's 7.8.0.

kklamberty (approved):

PR #1504, opened 2024-03-05
Fix deployment issue

NicMcPhee:
When I build the project on the droplet, all I ever get is 404 errors. I ended up doing a complex version of `git bisect` only working on PR merges, and finally narrowed it down to #1453 ("Deal with large bundle sizes"). As well as upping the budget for bundle sizes, that PR changed from using the `browser` builder to the `application` builder. That's worked fine in development, but seems to fail in deployment.

This should fix the deployment problem. I'm not honestly sure why changing the builder matters, but based on way too much fiddling on the droplet, I think this will fix problem.

### DIFF
--- a/client/angular.json
+++ b/client/angular.json
@@ -15,11 +15,11 @@
       "prefix": "app",
       "architect": {
         "build": {
-          "builder": "@angular-devkit/build-angular:application",
+          "builder": "@angular-devkit/build-angular:browser",
           "options": {
             "outputPath": "dist/client",
             "index": "src/index.html",
-            "browser": "src/main.ts",
+            "main": "src/main.ts",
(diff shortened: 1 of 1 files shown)


NicMcPhee:
It did fix the problem. Doing some more reading, we totally need to set the `"builder"` property to `"@angular-devkit/build-angular:browser"`. That tells it that we want to:

> Build an Angular application targeting a browser environment using [Webpack](https://webpack.js.org/).

which is what we want here. Setting it to `application` says we want to:

> Build an Angular application targeting a browser and server environment using [esbuild](https://esbuild.github.io/).

This seems like it could be a plausible target for us, but we're not using an Angular server environment which may be why that didn't work.

PR #1580, opened 2024-08-16
Add `cy.intercept()`/`cy.wait()` to stop E2E failures due to timing

### LINKED ISSUE #649: Cypress intermittently fails because of detached elements in the DOM
I haven't looked into how to fix this issue—we may need to adjust our retry logic? 🤷🏻‍♂️

[Here](https://drive.google.com/file/d/1Cf8COuoPOCdm4OPkB8fljIcs031uvoRw/view?usp=drivesdk) are logs from a failing run. The relevant section is this:

```
2022-01-23T23:49:24.6126996Z [31m  1 failing[0m
2022-01-23T23:49:24.6127177Z 
2022-01-23T23:49:24.6127500Z [0m  1) User list
2022-01-23T23:49:24.6127979Z        Should type something in the age filter and check that it returned correct elements:
2022-01-23T23:49:24.6128931Z [0m[31m     CypressError: Timed out retrying after 4000ms: `cy.should()` failed because this element is detached from the DOM.
2022-01-23T23:49:24.6129248Z 
2022-01-23T23:49:24.6130007Z `<mat-card-title _ngcontent-pqi-c75="" class="mat-card-title user-card-name">Bolton ...</mat-card-title>`
2022-01-23T23:49:24.6130357Z 
2022-01-23T23:49:24.6130818Z Cypress requires elements be attached in the DOM to interact with them.
2022-01-23T23:49:24.6131073Z 
2022-01-23T23:49:24.6131772Z The previous command that ran was:
2022-01-23T23:49:24.6131979Z 
2022-01-23T23:49:24.6132227Z   > `cy.get()`
2022-01-23T23:49:24.6132387Z 
2022-01-23T23:49:24.6132898Z This DOM element likely became detached somewhere between the previous and current command.
2022-01-23T23:49:24.6133197Z 
2022-01-23T23:49:24.6133539Z Common situations why this happens:
2022-01-23T23:49:24.6134058Z   - Your JS framework re-rendered asynchronously
2022-01-23T23:49:24.6134647Z   - Your app code reacted to an event firing and removed the element
2022-01-23T23:49:24.6134899Z 
2022-01-23T23:49:24.6135596Z You typically need to re-query for the element or add 'guards' which delay Cypress from running new commands.
2022-01-23T23:49:24.6135907Z 
2022-01-23T23:49:24.6136460Z https://on.cypress.io/element-has-detached-from-dom[0m[90m
```

(Apologies for the ANSI escape noise. For best viewing, run `less -R <logfile>`.)

### LINKED ISSUE #1319: Cypress test race condition in Add-User.spec.ts
This was commented on in the lab today, but when you run the cypress tests for the add-user off the template, you get a race-condition on getting the email component. The specific error is referring to trying to type into a disabled element.

NicMcPhee:
I think this will ensure that the `page.addUser(user)` calls actually complete before we start looking for page elements. Hopefully that will reduce/eliminate the problems we've been having with E2E tests failing sporadically, presumably due to timing issues.

I also added a 10 second timeout to `getSnackBar` (along with a comment about why). This hopefully helps ensure that we actually _have_ a snackbar visible before we start asserting things about it (which will fail if it's not yet there). (We can't use the `cy.intercept()`/`cy.wait()` technique because failure doesn't route us to a new page.)

While I was here I fixed/improved our `addUser` error messages, which weren't actually correct in `AddUserComponent`. We were saying that the server failed, for example, in cases where we passed the server an illegal `User` to add. This now explicitly checks for error codes 400 and 500, as well as having a fall-back error for other codes.

We could (and maybe should?) add `intercept()/wait()` pairs around our other network calls? Those haven't tended to time out in the past, so I'm not sure. We might be able to put those in our `po.ts` files so we don't have to repeat them everywhere.

Hopefully this resolves two long-standing issues:

- Closes #1319
- Closes #649


client/cypress/e2e/add-user.cy.ts
```diff
@@ -118,10 +141,15 @@ describe('Add user', () => {
         role: 'editor'
       };
 
+      // Here's we're _not_ expecting to route to `/api/users` since adding this
```
kklamberty, line 144:
```suggestion
      // Here we're _not_ expecting to route to `/api/users` since adding this
```

client/cypress/e2e/add-user.cy.ts
```diff
@@ -90,13 +90,36 @@ describe('Add user', () => {
         role: 'editor'
       };
 
+      // The `page.addUser(user)` call ends with clicking the "Add User"
+      // button on the interface. That then leads to the client sending an
+      // HTTP request to server, which is has to process that request
```
kklamberty, line 95:
```suggestion
      // HTTP request to the server, which has to process that request
```

client/cypress/e2e/add-user.cy.ts
```diff
@@ -90,13 +90,36 @@ describe('Add user', () => {
...
+      // (including making calls to add the user to the database and wait
+      // for those to respond) before we get a response and can update the GUI.
+      // By calling `cy.intercept()` we're saying we want Cypress to "notice"
+      // when we go to `/api/users`. The `AddUserComponent.submitForm()` method
+      // routes to `/api/users/{MongoDB-ID}` if the REST request to add the user
+      // succeeds, and that routing will get "noticed" by the Cypress because
+      // of the `cy.intercept()` call.
+      //
+      // The `.as('addUser')` call basically gives that event a name (`addUser`)
+      // which we can use in things like `cy.wait()` to say which event or events
```
kklamberty, line 105: What does it look like to wait for more than one event?

client/.nx/cache/nx_files.nxt
kklamberty, line 1: Should this file have been in the .gitignore?

client/cypress/e2e/add-user.cy.ts
```diff
@@ -118,10 +141,15 @@ describe('Add user', () => {
...
+      // Here's we're _not_ expecting to route to `/api/users` since adding this
+      // user should fail. So we don't add `cy.intercept()` and `cy.wait()` calls
+      // around this `page.addUser(user)` call. If we _did_ add them, the test wouldn't
+      // actually fail because a `cy.wait()` that times out isn't considered a failure,
+      // although we could catch the timeout and turn it into a failure if we needed to.
       page.addUser(user);
 
       // We should get an error message
-      page.getSnackBar().should('contain', `Problem contacting the server – Error Code:`);
+      page.getSnackBar().should('contain', 'Tried to add an illegal new user');
```
kklamberty, line 152: Good catch!

client/src/app/users/add-user.component.ts
kklamberty, line 1: Nicely handled!

client/src/app/users/user.service.ts
```diff
@@ -119,6 +119,7 @@ export class UserService {
 
   addUser(newUser: Partial<User>): Observable<string> {
     // Send post request to add a new user with the user data as the body.
+    // `res.id` should be the MongoDB ID of the newly added `User`.
```
kklamberty, line 122: Should we say anything about res being the response? Or, should the name of the thing just be changed to `response` instead of `res` to help clarify?

kklamberty (changes requested):
I left a couple of minor edits to the comments that I think should be fixed before merging in the PR, but overall, this looks great to me!

client/.nx/cache/nx_files.nxt
NicMcPhee: Good catch. Some research suggests that it should. Thanks!

client/cypress/e2e/add-user.cy.ts
NicMcPhee: You'd probably need to look at the docs where they have examples of this. I think an example would be a complex page that makes several requests to assemble its content (e.g., the current blog post, a list of titles of recent posts, and some ads), and you might want/need to wait until all those requests completed before you start checking the state of the page.

client/src/app/users/user.service.ts
NicMcPhee: I like the idea of changing the name to `response`.

NicMcPhee:
Thanks for the feedback and suggestions @kklamberty! I'm pretty hopeful that this will really help; I haven't had a single E2E failure since adding these changes.

Issue #1581, opened 2024-08-18
Remove uses of deprecated `RouterTestingModule`


NicMcPhee:
This has been deprecated, and might be removed entirely in Angular 19, so we should definitely remove our dependency on it.

I _think_ our primary dependency on it is in `UserProfileComponent`, which gets the ID of the user to display from the route, i.e., we go to `/users/1234` to display the profile for the user with ID 1234, and `UserProfileComponent` gets the ID from the `paramMap` of the URL.

I'm not entirely sure how to do this, TBH. One option I'd seen is creating a new `UserIdService` that provides the ID to the `UserProfileComponent` instead of having the `UserProfileComponent` extract it from the route/URL directly. That works best (I think) if there's a parent component that's watching for changes to the route/URL. It then extracts that ID and sends it to the service, and then routes to the `UserProfileComponent`. I'm a little worried that this just moves the lump to another part of the rug.

Another option (which might be simpler in the long run) is to make `UserProfileComponent` a _child_ component of whoever uses it. At the moment its used in two places:

- `UserCardComponent` (which is a child of `UserListComponent`) uses it to display the profile. We could perhaps change that to display in a pop-up view "over" the parent component(s)?
- `AddUserComponent` which is a stand-alone component. We could insert the profile as a chile component below the form for creating new users.

kklamberty:
Is it meaningful that bumping to Angular 19 still passes tests? https://github.com/UMM-CSci-3601/3601-iteration-template/pull/1643

NicMcPhee:
I think it's meaningful in the sense that they haven't pulled the plug on us yet, but I still think we want to fix this because eventually they might make this a breaking issue. So thanks for the work on #1648!

PR #1582, opened 2024-08-18
Replace construct-based injection with `inject()` calls

### LINKED ISSUE #1416: Switch from constructor-based injection to using explicit `inject()` calls
Having things like services injected into components via the constructor is awkward to look at and has often confused students.

There is apparently an explicit `inject()` function that we can use instead, and which seems to be where a lot of people in the Angular space are moving.

So instead of injecting services via the constructor, we can use something like:

```typescript
userListService = inject(UserListService);
```

If this actually works as cleanly as it sounds, we should totally move in this direction IMO.

NicMcPhee:
This replaces _most_ of the construct-based injection with `inject()` calls. The few remaining injectors are tied up with `ActivatedRoute` and the now deprecated `RouterTestingModule`. I think those can't be fixed without addressing that issue (#1581).

This makes progress on #1416, but doesn't entirely finish it.

### DIFF
--- a/client/src/app/company-list/company-list.component.ts
+++ b/client/src/app/company-list/company-list.component.ts
@@ -1,4 +1,4 @@
-import { Component, Signal } from '@angular/core';
+import { Component, inject, Signal } from '@angular/core';
 import { toSignal } from '@angular/core/rxjs-interop';
 import { CompanyCardComponent } from '../company-card/company-card.component';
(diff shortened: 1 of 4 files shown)


NicMcPhee:
All the updates from #1587 really changed the world this was based on, so I'm closing this one and starting over.

PR #1587, opened 2024-08-31
refactor: 4 suggestions from Angular nerd alum

michael-small:
Hey, it may be too late for this semester, but there was some changes I wanted to suggest. I think if it is possible to get these in before everyone is forking it, it would really establish some better patterns. I recently got promoted to lead frontend developer where I work and I am getting further involved in the community, and I want to give back some tips. 

~~(disclaimer - can't set up Java at the moment on my PC, so I wasn't able to load in real data, but this should be legit)~~ edit: WSL has been defeated, I can now run both Java and Mongo

Across the four commits bodies, I left messages about my reasoning. The four main takeaways

- Material can be brittle: target your own classes or the material component selectors, not `.mat-` or `.mdc-` classes. 
    - I am working through a monorepo with dozens of apps that had this issue. 
    - Material is getting better over time with customizations. 
    - The body of the comments give links to examples I made for people online to see the above points demonstrated, both problem and solution.
- Declaratively declare state
    - Just about anything done in a `constructor` or `ngOnInit` to initialize state can be done directly in the given variable. The second commit does `toSignal` on something directly on where the variable is declared. I do this all the time and it is really nice.
    - There are other examples akin to this in the repo (just about any observable can be piped/mapped/switchmapped right as it is declared) if you are interested in a longer term suggestion
- Replace manual unsubscribes with `takeUntilDestroy(this.destroyRef)`
    - Saves many lines and a lifecycle hook
- Convert a function to a pure function
    - Easier to track down state changes, goes hand in hand with `pipe/computed` values once in the habit. 

PS - the ngxtension dependency? I have a util in the library (form events, aka value/status/touched/pristine as one observable or signal), it just doesn't have a doc page yet lol. Though I also am doing a spike for an interactive doc for ngxtension that I could link as well.

### DIFF
--- a/client/src/app/company-list/company-list.component.ts
+++ b/client/src/app/company-list/company-list.component.ts
@@ -12,9 +12,7 @@ import { Company } from './company';
   styleUrl: './company-list.component.scss'
 })
 export class CompanyListComponent {
-  companies: Signal<Company[]>;
+  companies: Signal<Company[]> = toSignal(this.userService.getCompanies());
 
-  constructor(private userService: UserService) {
-    this.companies = toSignal(this.userService.getCompanies());
-  }
+  constructor(private userService: UserService) {}
 }
--- a/client/src/app/users/user-list.component.ts
+++ b/client/src/app/users/user-list.component.ts
@@ -108,8 +108,9 @@ export class UserListComponent implements OnInit {
    * get an updated list of `filteredUsers`.
    */
   public updateFilter(): User[] {
-     return this.userService.filterUsers(this.serverFilteredUsers, { name: this.userName, company: this.userCompany });
-  }
+    this.filteredUsers = this.userService.filterUsers(this.serverFilteredUsers, { name: this.userName, company: this.userCompany });
+    return this.filteredUsers;
+ }
 
   /**
    * Starts an asynchronous operation to update the users list
--- a/client/src/app/users/user-list.component.ts
+++ b/client/src/app/users/user-list.component.ts
@@ -30,74 +30,118 @@ import { toObservable, toSignal } from '@angular/core/rxjs-interop';
  * makes the most sense to do the filtering.
  */
 @Component({
-    selector: 'app-user-list-component',
-    templateUrl: 'user-list.component.html',
-    styleUrls: ['./user-list.component.scss'],
-    providers: [],
-    standalone: true,
-    imports: [AsyncPipe, MatCardModule, MatFormFieldModule, MatInputModule, FormsModule, MatSelectModule, MatOptionModule, MatRadioModule, UserCardComponent, MatListModule, RouterLink, MatButtonModule, MatTooltipModule, MatIconModule]
+  selector: 'app-user-list-component',
+  templateUrl: 'user-list.component.html',
+  styleUrls: ['./user-list.component.scss'],
+  providers: [],
+  standalone: true,
+  imports: [
+    AsyncPipe,
+    MatCardModule,
+    MatFormFieldModule,
+    MatInputModule,
+    FormsModule,
+    MatSelectModule,
+    MatOptionModule,
+    MatRadioModule,
+    UserCardComponent,
+    MatListModule,
+    RouterLink,
+    MatButtonModule,
+    MatTooltipModule,
+    MatIconModule,
+  ],
 })
 export class UserListComponent {
-    userName = signal<string | undefined>(undefined)
-    userAge = signal<number | undefined>(undefined)
-    userRole = signal<UserRole | undefined>(undefined)
-    userCompany = signal<string | undefined>(undefined)
+  userName = signal<string | undefined>(undefined);
+  userAge = signal<number | undefined>(undefined);
+  userRole = signal<UserRole | undefined>(undefined);
+  userCompany = signal<string | undefined>(undefined);
 
-    viewType = signal<'card' | 'list'>('card');
+  viewType = signal<'card' | 'list'>('card');
 
-    errMsg = signal<string | undefined>(undefined);
+  errMsg = signal<string | undefined>(undefined);
 
-    /**
-     * This constructor injects both an instance of `UserService`
-     * and an instance of `MatSnackBar` into this component.
-     * `UserService` lets us interact with the server.
-     *
-     * @param userService the `UserService` used to get users from the server
-     * @param snackBar the `MatSnackBar` used to display feedback
-     */
-    constructor(private userService: UserService, private snackBar: MatSnackBar) {
-        // Nothing here – everything is in the injection parameters.
-    }
+  /**
+   * This constructor injects both an instance of `UserService`
+   * and an instance of `MatSnackBar` into this component.
+   * `UserService` lets us interact with the server.
+   *
+   * @param userService the `UserService` used to get users from the server
+   * @param snackBar the `MatSnackBar` used to display feedback
+   */
+  constructor(private userService: UserService, private snackBar: MatSnackBar) {
+    // Nothing here – everything is in the injection parameters.
+  }
 
-    // Observable stuff needs observables to react to - just `toObservable` what is needed
-    private userRole$ = toObservable(this.userRole)
-    private userAge$ = toObservable(this.userAge)
+  // The `Observable`s used in the definition of `serverFilteredUsers` below need
+  // observables to react to, i.e., they need to know what kinds of changes to respond to.
+  // We want to do the age and role filtering on the server side, so if either of those
+  // text fields change we want to re-run the filtering. That means we have to convert both
+  // of those _signals_ to _observables_ using `toObservable()`. Those are then used in the
+  // definition of `serverFilteredUsers` below to trigger updates to the `Observable` there.
+  private userRole$ = toObservable(this.userRole);
+  private userAge$ = toObservable(this.userAge);
 
-    // We ultimately `toSignal` this to be able to access it synchronously, but we do all the RXJS operations internally
-    //     Once there is a value for both the role and age, the latest values will move onto the switchMap
-    serverFilteredUsers = toSignal(combineLatest([this.userRole$, this.userAge$]).pipe(
-        // You are now switching into another observable and mapping the previous values into the new one's args
-        switchMap(([role, age]) => this.userService.getUsers({
+  // We ultimately `toSignal` this to be able to access it synchronously, but we do all the RXJS operations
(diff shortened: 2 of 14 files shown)


client/src/app/company-list/company-list.component.ts
```diff
@@ -12,9 +12,7 @@ import { Company } from './company';
   styleUrl: './company-list.component.scss'
 })
 export class CompanyListComponent {
-  companies: Signal<Company[]>;
+  companies: Signal<Company[]> = toSignal(this.userService.getCompanies());
```
NicMcPhee, line 15: Nice – thanks!

NicMcPhee (changes requested):
Thanks so much for doing this! I don't think it's too late for this semester, and these are some nice changes. Unfortunately something in the last commit broke the filtering, although I'm not sure what yet.

I would also like to talk to you sometime about monorepos. I feel that we'd be a lot better off if we had a single monorepo to update instead of 5 repos to keep up-to-date and in sync, but I've never really taken the time to figure out what that would look like.

client/src/app/users/user-list.component.ts
```diff
@@ -108,9 +107,8 @@ export class UserListComponent implements OnInit, OnDestroy  {
    * Called when the filtering information is changed in the GUI so we can
    * get an updated list of `filteredUsers`.
    */
-  public updateFilter(): void {
-    this.filteredUsers = this.userService.filterUsers(
-      this.serverFilteredUsers, { name: this.userName, company: this.userCompany });
+  public updateFilter(): User[] {
+     return this.userService.filterUsers(this.serverFilteredUsers, { name: this.userName, company: this.userCompany });
```
NicMcPhee, line 111: This is the problem. The HTML for this component calls `updateFilter()` whenever there is a change in the input fields. It then expects that to change the value of `this.filteredUsers`, leading to a change in the displayed users. By making this more functional, we lost that expected side-effect, so typing in the text boxes has no effect.

If we change this to:

```typescript
  public updateFilter(): User[] {
     this.filteredUsers = this.userService.filterUsers(this.serverFilteredUsers, { name: this.userName, company: this.userCompany });
     return this.filteredUsers;
  }
```
everything works again.

This does raise the question, though, of whether there's a way to avoid this use of state/side-effects. We could have the HTML call `userService.filterUsers()` every time it wants a filtered list of users?

michael-small: Ah, yeah I see the issue now. Sorry, I would have realized this quick if I could run it. Having issues getting Mongo working with my WSL configuration.

I can make that change above to get it working if you want. 

Otherwise, check out this commit in the alt branch: https://github.com/UMM-CSci-3601/3601-iteration-template/commit/8f636136b87bf87eff3edceca79789398ef40976

`filteredUsers` would then be a `computed` value. Whenever the `serverFilteredUsers` changes or `userName` or `userCompany` changes, the filter would fire off and update the computed value.

michael-small: I pushed up your suggested change

michael-small:
> I would also like to talk to you sometime about monorepos. I feel that we'd be a lot better off if we had a single monorepo to update instead of 5 repos to keep up-to-date and in sync, but I've never really taken the time to figure out what that would look like.

Ah, like each group's project in a single monorepo?

client/src/app/users/user-list.component.ts
michael-small: The signal ngModel binding in that linked branch is now supported, and those relevant signals firing would avoid the need for calling a function like this new version that has side effects

client/src/app/users/user-list.component.ts
```diff
@@ -76,15 +75,15 @@ export class UserListComponent implements OnInit, OnDestroy  {
...
+      takeUntilDestroyed(this.destroyRef)
     ).subscribe({
       // Next time we see a change in the Observable<User[]>,
       // refer to that User[] as returnedUsers here and do the steps in the {}
       next: (returnedUsers) => {
         // First, update the array of serverFilteredUsers to be the User[] in the observable
         this.serverFilteredUsers = returnedUsers;
         // Then update the filters for our client-side filtering as described in this method
-        this.updateFilter();
+        this.filteredUsers =  this.updateFilter();
```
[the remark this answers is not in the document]
michael-small: Hoping my latest push resolves this

client/src/app/users/user-list.component.ts
michael-small: This all said, not sure if that would actually work without getting Mongo running. If that approach did work, IMO that is the best way to not have side effects.

michael-small:
Ok, sorry I'm all over the place in this PR with responses but I am excited that I got things working locally. WSL is a frankenstein mess but when the monster works it works.

What I would suggest to fulfill your question about reducing side effects of the client side filtering: compute `filteredUsers` as a `computed` signal. How so? Check out this branch I made right off of this one: https://github.com/UMM-CSci-3601/3601-iteration-template/compare/main...michael-small:3601-iteration-template:feat-fall24-alum-suggestions-filter-via-computed

If you like this, I can merge that branch into this one

NicMcPhee:
> > I would also like to talk to you sometime about monorepos. I feel that we'd be a lot better off if we had a single monorepo to update instead of 5 repos to keep up-to-date and in sync, but I've never really taken the time to figure out what that would look like.
> 
> Ah, like each group's project in a single monorepo?

Actually, that's not what I was thinking of. I was just thinking of managing our "started" repos in this organization. Currently we have multiple labs and this iteration template all in this organization, and keeping them all synced up and up-to-date is a real pain. The changes you're proposing, for example, are really good, but then need to be copied over to Labs 3 and 4. And when there's an Angular or Javalin update we have to make those changes in multiple repositories, which is a pain.

I'm not sure monorepos would make sense for the _students_ because the groups keep changing so it's unlikely that there would be a single student's universe that groups would want to work on. I hadn't really thought about that, though, so maybe there are advantages and ways to make that work that I hadn't thought of.

NicMcPhee:
> What I would suggest to fulfill your question about reducing side effects of the client side filtering: compute `filteredUsers` as a `computed` signal. How so? Check out this branch I made right off of this one: [main...michael-small:3601-iteration-template:feat-fall24-alum-suggestions-filter-via-computed](https://github.com/UMM-CSci-3601/3601-iteration-template/compare/main...michael-small:3601-iteration-template:feat-fall24-alum-suggestions-filter-via-computed)
> 
> * If you like this, I can merge that branch into this one
> * I can also comment back in the original comments from the file

Of course, computed signals would be exactly the way to do. Thanks!

I'd love to have you commit all this back in here if you're OK with that. These are really nice changes and would really improve the code, and I really appreciate your contributions!

michael-small:
Merged the computed in

michael-small:
For reference, you can also do this in the user profile if you wanted. In this PR I remove an uneeded `this` in the template (and `this` will be unsupported in templates soon)

```ts
    // `map` and `switchMap` are RXJS operators, and
    // each represents a step in the pipeline built using the RXJS `pipe`
    // operator.
    // The map step takes the `ParamMap` from the `ActivatedRoute`, which
    // is typically the URL in the browser bar.
    // The result from the map step is the `id` string for the requested
    // `User`.
    // That ID string gets passed (by `pipe`) to `switchMap`, which transforms
    // it into an Observable<User>, i.e., all the (zero or one) `User`s
    // that have that ID.
    user = toSignal(
        this.route.paramMap.pipe(
            // Map the paramMap into the id
            map((paramMap: ParamMap) => paramMap.get('id')),
            // Maps the `id` string into the Observable<User>,
            // which will emit zero or one values depending on whether there is a
            // `User` with that ID.
            switchMap((id: string) => this.userService.getUserById(id)),
            catchError(_err => {
                this.error.set( {
                    help: 'There was a problem loading the user – try again.',
                    httpResponse: _err.message,
                    message: _err.error?.title,
                })
                return of<User>()
            }),
            /*
            * You can uncomment the line that starts with `finalize` below to use that console message
            * as a way of verifying that this subscription is completing.
            * We removed it since we were not doing anything interesting on completion
            * and didn't want to clutter the console log
            */
            // finalize(() => console.log('We got a new user, and we are done!'))
        ),
    )
```

NicMcPhee:
> For reference, you can also do this in the user profile if you wanted. In this PR I remove an uneeded `this` in the template (and `this` will be unsupported in templates soon)

This is awesome! I've tried doing this several times, and never managed to make it work. I think the thing I was missing was `catchError`, which I didn't know existed. (I'm sure it's "obvious" if you read the right documentation, but it wasn't something I ever ran across.)

michael-small:
> > For reference, you can also do this in the user profile if you wanted. In this PR I remove an uneeded `this` in the template (and `this` will be unsupported in templates soon)
> 
> This is awesome! I've tried doing this several times, and never managed to make it work. I think the thing I was missing was `catchError`, which I didn't know existed. (I'm sure it's "obvious" if you read the right documentation, but it wasn't something I ever ran across.)

Yeah, it can be a little tricky to understand the whole `of` thing (I just edited in a comment to the block), and if you are mapping the observable and the `catchError` is not in the right hierarchy of the pipe, but it is what people tend to do these days. And doing `finalize` rather than `complete`.

NicMcPhee:
Some of the tests fail because of the changes in the structures, and I'm working on fixing those now. It's possibly that my changes and your changes might smash into each other. Should I step away and let you finish the changes to the "production" code?

michael-small:
> Some of the tests fail because of the changes in the structures, and I'm working on fixing those now. It's possibly that my changes and your changes might smash into each other. Should I step away and let you finish the changes to the "production" code?

I'll remove the `this` in template and then stop putting in more changes

NicMcPhee:
Thanks so much for all your great suggestions! This is where I was hoping we'd eventually get to, but I kept getting stuck on little things. Your expertise is really helping improve things a ton!

michael-small:
Happy to help. And I see what you mean by the changes to the structure, with all the primitive/object values to signals of those, and pulling out hooks. Thanks for cleaning that up after I broke those tests, you should be clear to go now. I pushed up removing the remaining `this` in templates and I won't add more changes. Thanks for taking in this feedback, I am very grateful that 3601 works like it does and got me started in Angular.

NicMcPhee:
Students and alums like you have contributed a lot over the years, and deserve a lot of the credit. I look forward to mentioning your contributions in class!

NicMcPhee (dismissed):
Thanks to @michael-small for all the awesome improvements!

NicMcPhee (approved):

PR #1598, opened 2024-09-02
Add a 30 sec timeout to seeding the DB in Cypress tests

NicMcPhee:
This adds a 30 second timeout to the DB seeding in all the relevant Cypress tests. I'm hoping this will reduce the flakiness of our E2E tests.


NicMcPhee:
I'm not convinced this works, or was a good idea, so I'm closing it.

PR #1646, opened 2025-01-10
Fix config issues in `angular.json`

### LINKED ISSUE #1623: Fix the `angular.json` configuration issues
Somehow the projects last semester included fixes to `angular.json` that were never copied back into the iteration template, so I ended up having to debug them all over again. We really need to fix that so it doesn't happen again.

See [this PR](https://github.com/UMM-CSci-3601-F24/it-2-virgin-of-the-passion/pull/38) for the details of what needs to changed. The changes to the budget limits aren't necessary, but are probably a good idea just to keep the noise down.

kklamberty:
Closes #1623


NicMcPhee:
I think this is undoing something I intentionally did last year. See [this commit](https://github.com/UMM-CSci-3601/3601-iteration-template/commit/9bdb77aa95be4a8522c69c3a121df4ef93ed48cb) for some limited notes. Feel free to holler if you want to chat about this.

kklamberty:
> I think this is undoing something I intentionally did last year. See [this commit](https://github.com/UMM-CSci-3601/3601-iteration-template/commit/9bdb77aa95be4a8522c69c3a121df4ef93ed48cb) for some limited notes. Feel free to holler if you want to chat about this.

If that's the case, you may not have been happy with that change according to the issue you created several months later:
https://github.com/UMM-CSci-3601/3601-iteration-template/issues/1623#issue-2633556126

I cherry-picked this commit based on what you wrote in that issue. It sounds like something that caused you to need to chase down a problem. I noticed that when I upgraded to Angular 19, one of the optional migrations was to move to a new builder, but I read the notes about it and looked at what builder we had (after I did this commit in my fork based on issue #1623), it was `browser`, so, I intentionally did not change it to anything else. 

Looking at iteration 3, it looks like you may have had the class cherry-pick that commit too. I don't think you were happy with it.

NicMcPhee:
All that's entirely possible. All I knew was that I'd done stuff with the builder business. It looks like you've actually dug into this way more than I did (or remember doing). So maybe we do want to undo what I did?

kklamberty:
It seems like you already did undo it, but just not here yet. So, yes, I think it makes sense. The issue you wrote up explains a little more, and the commit in this PR is yours. It sounds like it's something we might not notice until deployment, but I'm not certain. The work, debugging, and rework and re-debugging were yours, and I'm happy to avoid needing to figure this out for a third time this semester. I am comfortable with this PR being merged.

NicMcPhee (approved):
Sorry for breaking stuff, and thanks for fixing it!

PR #1648, opened 2025-01-12
1581 remove uses of deprecated routertestingmodule

### LINKED ISSUE #1581: Remove uses of deprecated `RouterTestingModule`

kklamberty:
I was able to replace the use of RouterTestingModule following the instructions in the deprecation message:

@deprecated
Use provideRouter or RouterModule/RouterModule.forRoot instead. This module was previously used to provide a helpful collection of test fakes, most notably those for Location and LocationStrategy. These are generally not required anymore, as MockPlatformLocation is provided in TestBed by default. However, you can use them directly with provideLocationMocks.

combined with the useful posts here: https://github.com/angular/angular/issues/54853 and particularly the example mentioned here: https://github.com/angular/angular/issues/54853#issuecomment-1997566987

This branch includes two cherry-picked commits from a fork of this repo:
- one that replaced one instance of RouterTestingModule: https://github.com/kklamberty/3601-iteration-template/commit/5fb17667ffe3479ea128efc7e4e531786156ad48
- one that replaces the remaining instances: https://github.com/kklamberty/3601-iteration-template/commit/7d2483ddd354dc096a76bdbf599c9e42e5125a88

Both commits relate to an issue in the repo from which this fork came: https://github.com/UMM-CSci-3601/3601-iteration-template/issues/1581 and I think that accepting this pull request closes #1581


kklamberty:
I'm not totally sure *why* the Misbehaving User List test fails in ways that it didn't before, but one of the changes in this PR helps make other dependencies happy. I would like to see this PR merged so that other branches and PRs from dependabot might start working again.

client/src/app/users/add-user.component.spec.ts
```diff
@@ -282,7 +281,7 @@ describe('AddUserComponent#submitForm()', () => {
         MatSelectModule,
         MatInputModule,
         BrowserAnimationsModule,
-        RouterTestingModule.withRoutes([
+        RouterModule.forRoot([
```
NicMcPhee, line 284: Oh, nice! This turned out to be quite easy to deal with, where I had worried that it was going to be a big complicated thing. Well done!

NicMcPhee (approved):
Thanks for figuring this out! I definitely think it does resolve the issue that you mentioned in the write-up, so definitely feel free to close that as well.

PR #1664, opened 2025-01-21
Update eslint

kklamberty:
ESLint 8 has reached end of life, and it looks like it's important to update this to version 9 or more. Luckily, @NicMcPhee already did some work to modernize our eslint configuration, so I was able to basically just follow the migration guide here: https://eslint.org/docs/latest/use/migrate-to-9.0.0

However, I'm still not sure I have completed this task, nor do I know how to check. I *think* that the name of the file generated ending in .mjs might mean I need to do something more (at least rename that file to just .js at the end?) but I don't know. When I do rename the file to end in just `.js` rather than `.mjs`, I get some additional warning:
> (node:34240) [MODULE_TYPELESS_PACKAGE_JSON] Warning: Module type of file:///Users/lamberty/Documents/3601/3601-iteration-template/client/eslint.config.js?mtime=1737432342100 is not specified and it doesn't parse as CommonJS.
Reparsing as ES module because module syntax was detected. This incurs a performance overhead.
To eliminate this warning, add "type": "module" to /Users/lamberty/Documents/3601/3601-iteration-template/client/package.json.
(Use `node --trace-warnings ...` to show where the warning was created)

I don't know if that warning tells me more (like, hey, you're actually linting now), or if leaving it as mjs is preferred. Also, the file contains more `...` than I would have expected. I don't know what those mean. I see references about it being the spread operator and about it being used to indicate things that are overwritten. The overwriting idea makes some sense in this context... like you will be given some default setting, and instead you will use this. But, I'd still like to have input from others on this PR in case I've done something wildly incorrect or I've somehow turned off linting. Any thoughts?

I installed a node deprecation checker (`npm install npm-deprecated-check`) and used `ndc current` to check the installation, and that made it seem like I have maybe at least succeeded in removing the deprecated eslint.


NicMcPhee (approved):
I think the `.mjs` is just a a JS _module_ file, taking advantage of some newer JS features. It basically just moved the rules in `.eslintrc.json` into that JS module file.

The `...` is similar to the spread operator in Python if you've used that. I don't know if I've ever used it in JS/TS, but we could potentially file it useful in places?

It would probably make sense to sit down and go over (and document) the contents of this file to make it easier to understand and maintain downstream.

PR #1674, opened 2025-02-05
Manual angular update using cli

kklamberty:
Rather than using the pull requests brought up separately by dependabot and the updater for Angular, I ran `ng update` and checked which things would be updated. They all seemed to match the PRs from the bots, so I think I captured all the right updates. If I did, those PRs will close automatically.

In the process of running the tests, I noticed that there were several warnings coming up about elements in the style sheets, and I thought that might make students nervous. I searched for [an answer on stack overflow](https://stackoverflow.com/a/64393470/3842582) and in the [SASS documentation](https://sass-lang.com/documentation/breaking-changes/color-functions/)  to figure out the simplest possible way to follow the given instructions. All of the component tests and E2E tests pass, and, furthermore, the pages look the same. There is still one warning about polyfills being present but unused, but I'm not as sure about changing that without looking more into deployment and making sure it's not important later.


mahon479 (approved):
I do not see any issues here. Looks good.

